PR #1, opened 2026-03-01
Add meerkat-net package with circuit relay support for distributed runtime.

ShreejithBabuG:
This PR adds the meerkat-net package to meerkat-lib/net as discussed.

Features:
- Actor-based network layer using Kameo and libp2p
- Circuit relay v2 for browser-to-browser communication
- NetworkLayer trait with real and mock implementations


JonathanAldrich (approved):

PR #4, opened 2026-03-18
Extend evaluator with Func and Action expressions

ShreejithBabuG:
Extends the evaluator with support for creating closures and action closures.

**Changes:**
- Implemented Expr::Func for creating regular closures
- Implemented Expr::Action for creating action closures
- Added unit tests: test_func_and_call, test_action_creation
- All 6 tests passing

**Testing:**
- test_func_and_call: verifies function creation and application
- test_action_creation: verifies action closure creation

All core expression types are now implemented (Literal, Variable, Call, Binop, Unop, If, Func, Action).

### DIFF
--- a/meerkat-lib/src/runtime/manager/mod.rs
+++ b/meerkat-lib/src/runtime/manager/mod.rs
@@ -1,7 +1,65 @@
-pub mod action;
-pub mod assert;
-pub mod handler;
-// pub mod init;
+use std::collections::HashMap;
+use super::ast::{Value, Decl};
+use super::interpreter::{eval, EvalError};
 
-mod manager_def;
-pub use manager_def::Manager;
+pub struct Service {
+    pub name: String,
+    pub vars: HashMap<String, Value>,
+}
+
+pub struct Manager {
+    pub services: HashMap<String, Service>,
+    pub current_service: Option<String>,
+}
+
+impl Manager {
+    pub fn new() -> Self {
+        Manager {
+            services: HashMap::new(),
+            current_service: None,
+        }
+    }
+
+    pub async fn create_service(&mut self, name: String, decls: Vec<Decl>)
+        -> Result<(), EvalError>
+    {
+        let mut service = Service {
+            name: name.clone(),
+            vars: HashMap::new(),
+        };
+        self.current_service = Some(name.clone());
+
+        for decl in decls {
+            match decl {
+                Decl::VarDecl { name, val } |
+                Decl::DefDecl { name, val, .. } => {
+                    let value = eval(&val, &vec![], self).await?;
+                    service.vars.insert(name, value);
+                }
+                Decl::TableDecl { .. } => {
+                    return Err(EvalError::NotImplemented);
+                }
+            }
+        }
+
+        self.services.insert(name.clone(), service);
+        Ok(())
+    }
+
+    pub async fn lookup(&mut self, ident: &str) -> Result<Value, EvalError> {
+        if let Some(service_name) = self.current_service.clone() {
+            if let Some(service) = self.services.get(&service_name) {
+                if let Some(value) = service.vars.get(ident) {
+                    return Ok(value.clone() as Value);
+                }
+            }
+        }
+        Err(EvalError::LookupError(format!("Variable '{}' not found", ident)))
+    }
+}
+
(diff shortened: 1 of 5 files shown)


JonathanAldrich (changes requested):
This generally looks good, however, I anticipate we will likely need to pass closures over the network.  So, we don't want them to be any larger than necessary.  The closures could be smaller if the stored environment included only the variables that are actually used (the "free variables").  Can you compute this and then copy only a subset of env into the closure?  You should be able to use free_var in runtime/semantic_analysis/typecheck/utils.rs for computing the free variables.

meerkat-lib/src/runtime/semantic_analysis/typecheck/tc_expr.rs
```diff
@@ -13,8 +13,7 @@ impl TypecheckEnv {
                 Value::String { .. } => String,
                 Value::Closure { .. } => panic!("Cannot typecheck closure literal"),
                 Value::ActionClosure { .. } => panic!("Cannot typecheck action closure literal"),
-            },
-            Expr::KeyVal { key, value } => self.infer_expr(&value),
+            },            Expr::KeyVal { key, value } => self.infer_expr(&value),
```
JonathanAldrich: Looks like a typo (deleted a newline by accident), can you revert this change?

meerkat-lib/src/runtime/semantic_analysis/typecheck/tc_expr.rs
```diff
@@ -209,45 +206,50 @@ impl TypecheckEnv {
...
+                    }
+                };
                 let func_type = self.infer_expr(operation);
                 let accum_type = self.infer_expr(identity);
-
                 match func_type {
                     Type::Fun(params, ret_type) => {
                         if !self.unify(&accum_type, &*ret_type) {
-                            panic!("Accumulator type should be the same as function return type, expected {}, got {}", &*ret_type, &accum_type);
+                            panic!("Accumulator type should match function return type");
```
JonathanAldrich: ok to replace "be the same as" with "match" but can we leave the expected/got with the actual types please?  helps us move toward better error messages

meerkat-lib/src/runtime/semantic_analysis/typecheck/tc_expr.rs
```diff
@@ -209,45 +206,50 @@ impl TypecheckEnv {
...
                         }
                         for param in params {
                             if !self.unify(&param, &column_type) {
-                                panic!("Column type should be the same as function argument type, expected {}, got {}", &param, &column_type);
+                                panic!("Column type should match function argument type");
```
JonathanAldrich: same as above

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -1,7 +1,129 @@
...
-mod manager_def;
-pub use manager_def::Manager;
+pub struct Service {
+    pub name: String,
+    pub vars: HashMap<String, Value>,
+}
+
+pub struct Manager {
+    pub services: HashMap<String, Service>,
+    pub current_service: Option<String>,
```
JonathanAldrich: This is going to be a problem--Manager won't be re-entrant if it requires a notion of a "current service."  And we need it to be re-entrant to (eventually) handle calls from one service to another.  So there should be no `current_service` field here. But, you need to know the right service to implement lookup. I think the solution is to pass the current service (or service name) to eval, and then have eval pass it to lookup.

JonathanAldrich: This is a problem--Manager can't have a notion of a "current service" because then it won't be re-entrant, and we need reentrancy to allow one service to call an operation of another.  So there should be no `current_service` variable here.  But you need to know the service to do lookup.  I think the solution is to pass a service (or service name) to `eval`, and have `eval` pass that to `lookup.`  This is kind of like `this` in OO languages.

JonathanAldrich (changes requested):
Thanks for adding the Manager implementation.  I found a couple of small issues and one somewhat larger one (regarding the current service), can you have a look at addressing these please?

JonathanAldrich (approved):
Looks good and tests pass--thanks for the PR!

PR #6, opened 2026-04-01
Implement def propagation and action execution

ShreejithBabuG:
- Defs are re-evaluated in topological order when a var is mutated
- Uses existing DependAnalysis infrastructure for dependency tracking
- Wired up main.rs to run services and @test blocks end-to-end
- Implemented free_var for ActionStmt in read_write.rs
- test0, test1, and test_func_update now passing end-to-end
- 1 new unit test verifying def updates after var change


JonathanAldrich (approved):

PR #21, opened 2026-04-14
Implement REPL for Meerkat (issue #13)

### LINKED ISSUE #13: Implement a REPL for Meerkat
The current Meerkat interpreter takes a file to run with a `-f` command line option.  But experimenting with code in a new language is easiest using a Read-Eval-Print Loop, or REPL for short.

Revise the main loop of the Meerkat interpreter so that when run without the `-f` option, it starts a REPL. The user should be able to enter one Meerkat definition or statement at a time.  If a definition or statement is incomplete after a line is entered, the REPL lets the user continue writing until it is complete.  Then the definition or statement is evaluated.  Definitions add bindings to the context that future commands and definitions can refer to.

The REPL should be able to evaluate all of the programs in the meerkat/tests directory that currently work with the `-f` option.  This can be tested either manually by entering lines one at a time or automatically by piping the test code to the meerkat interpreter.

Try to avoid duplication of code between the REPL and the existing parser and interpreter.  Use existing code wherever possible, and prefer to adapt existing code to make it more general rather than duplicating it or writing new code that is similar for the REPL.

Val Waeselynck [talks about REPLs and what makes a good one](https://vvvvalvalval.github.io/posts/what-makes-a-good-repl.html).  It's not necessary to implement all of the characteristics that Val mentions (some of which go well beyond what is in this issue), but look at it to get an idea of (A) how to design user-visible features that are part of this issue and (B) thinking about what we might want to add eventually, which might influence the design of the code.

nivethadhanakoti:
Closes #13

•⁠  ⁠⁠ cargo run ⁠ with no arguments now launches an interactive REPL
•⁠  ⁠Multi-line input is handled transparently — a continuation prompt ("       >") appears until the statement is complete
•⁠  ⁠Services defined in one REPL turn are visible to ⁠ @test ⁠ blocks entered in subsequent turns (shared manager across the loop)
•⁠  ⁠Works non-interactively: pipe any test file directly ⁠ cargo run < meerkat/tests/test1.meerkat ⁠)
•⁠  ⁠All existing test files pass via both ⁠ -f ⁠ and pipe mode
•⁠  ⁠Existing ⁠ -f ⁠ and ⁠ -s ⁠ behaviour is unchanged

### DIFF
--- a/meerkat-lib/src/runtime/parser/lex.rs
+++ b/meerkat-lib/src/runtime/parser/lex.rs
@@ -58,7 +58,7 @@ fn skip_multi_line_comments<'b>(lex: &mut Lexer<'b, Token<'b>>) -> Skip {
                         balanced_comments += 1;
                     }
                 }
-                None => panic!("Reached end of file"),
+                None => break,
                 _ => {
                     lex.bump_unchecked(1);
                 }
--- a/meerkat-lib/src/runtime/parser/mod.rs
+++ b/meerkat-lib/src/runtime/parser/mod.rs
@@ -7,11 +7,22 @@ pub mod meerkat {
     include!(concat!(env!("OUT_DIR"), "/runtime/parser/meerkat.rs"));
 }
 
+/// Result of attempting to parse a REPL input buffer.
+pub enum ReplParseResult {
+    /// Input parsed successfully into one or more statements.
+    Complete(Vec<crate::ast::Stmt>),
+    /// Input is syntactically incomplete (e.g. an open brace with no matching close).
+    /// The REPL should prompt for more input and append it to the buffer.
+    Incomplete,
+    /// Input has a real syntax error that won't be resolved by adding more text.
+    Error(String),
+}
+
 pub mod parser {
     use logos::Logos;
 
     use crate::ast::Stmt;
-    
+
     pub fn parse_string(input: &str) -> Result<Vec<Stmt>, String> {
         let lex_stream = super::lex::Token::lexer(input)
             .spanned()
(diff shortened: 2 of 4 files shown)


JonathanAldrich (changes requested):
In general this looks really great!  I noticed two small issues.

First, if a test file ends with a closing brace and then a newline, I get an error message:

```
aldrich@ALDRICH-2024:~/develop/meerkat$ cargo run < meerkat/tests/test0.meerkat
Service 'foo_service' loaded.

thread 'main' (347800) panicked at meerkat-lib/src/runtime/parser/lex.rs:61:25:
Reached end of file
aldrich@ALDRICH-2024:~/develop/meerkat$
```

Running the same file with -f works.  And running test1.meerkat with stdin redirection works, I guess because test1.meerkat ends without a newline.

Second, it looks like networking is not enabled in the repl.  For example, the following command fails:

```
aldrich@ALDRICH-2024:~/develop/meerkat$ cargo run -- -i /ip4/128.2.220.206/tcp/9000/p2p/12D3KooWSD9nkwNyF7LeUZDWv2L1oLYUpmAaqJeerU3D4952RjfJ/s1 < meerkat/tests/test_client.mkt
Remote service 's1' registered at /ip4/128.2.220.206/tcp/9000/p2p/12D3KooWSD9nkwNyF7LeUZDWv2L1oLYUpmAaqJeerU3D4952RjfJ/s1.
Error: Service 's2': Network error: No network layer available
Error: @test(s2): Lookup error: Variable 'w' not found in service 's2'
aldrich@ALDRICH-2024:~/develop/meerkat$
```

Can you look at addressing these?

JonathanAldrich (approved):
Looks good!

Issue #14, opened 2026-04-05
Add support for expression evaluate and `let` to the REPL


JonathanAldrich:
The current test programs (and the implementation of parsing) only supports definitions at the top level.  We would like to support statements and expression evaluation.  It should be possible to execute any statement that can appear in a `@test` block (e.g. do and assert statements).  It should also be possible to write an expression statement of the form 

`a+3;`

When an expression statement is evaluated, the value is printed by the REPL before prompting the user for the next input.  For example if `a` is `2` then the above would print `5`.

Finally, a `let` statement allows bindings to be added to the environment, which can be used later.  The following results in printing 5:

```
let a = 2;
a+3;
```

Let statements should also work in @test blocks (both in the REPL and when using `-f`).  It's OK for expression statements to appear in @test blocks, though they are not useful (arguably we should give a warning about this).

JonathanAldrich:
Closing this issue as it was addressed by PR #22

PR #22, opened 2026-04-16
Implement #14: add let and expression statements to REPL

nivethadhanakoti:
#14 

- `let` statements bind variables in the environment for later use
- Expression statements (`a + 3;`) evaluate and print their value in the REPL
- `let` bindings persist across REPL inputs
- `let` and expression statements work inside `@test` blocks with both REPL and `-f`
- `do` and `assert` statements work as top-level REPL input

Tests passing: all existing local tests pass, plus new REPL behaviour verified manually.


JonathanAldrich (approved):

Issue #15, opened 2026-04-05
Add `watch` to the REPL


JonathanAldrich:
Add a `watch` statement to the repl:

```
watch svc.x+1
```

Watch sets up a hidden `def` bound to the given expression.  Whenever the expression changes (e.g. because a transaction has modified `svc.x`) then the REPL will report that the watch expression changed and show its new value (please propose a nice output syntax for this).

To avoid interference between writing code and reporting updates, updates are reported only after each time the user enters a complete statement or definition, or presses <enter> to denote readiness to receive updates.  Note that update propagation is not instantaneous, so executing a transaction in the REPL may not result in immediately seeing all the updates it causes.

JonathanAldrich:
Implemented in #23

PR #23, opened 2026-04-24
Implement #15: add watch to the REPL

ShreejithBabuG:
- `watch <expr>;` registers an expression to monitor in the REPL
- Shows current value on registration: `Watching: foo.y (current value: 2)`
- After each complete input or empty enter, checks all watches and reports changes: `[watch] foo.y: 2 => 6`
- Works with service member access (e.g. `watch svc.x+1`), let bindings, and any valid expression
- Proposed output syntax: `[watch] <expr>: <old> => <new>` on change, `[watch] <expr> = <val>` on first registration

All existing tests pass.


JonathanAldrich (approved):
Looks good, thanks for the PR.  I noticed it works with watching defs in remote services too! (which was the intent, but I think not explicit in the issue)

PR #26, opened 2026-04-26
Fix #25: preserve service context when calling closures

### LINKED ISSUE #25: current service not captured properly in actions
It looks like the current service is not captured properly in actions in all circumstances.  In particular, consider the following entered in the REPL:

```
service act_fn_bug {
    var x = 0;
    pub def add_to_x = fn n => action { x = x + n; };
}

do act_fn_bug.add_to_x(3);
```

Executing this results in the error message:

`Error: Lookup error: Variable 'x' not found in service ''`

nivethadhanakoti:
Fixes #25

When a closure (e.g. fn n => action { x = x + n; }) was called from the REPL or any context with a different service, the body was evaluated with the caller's service context ("") instead of the closure's original service. This caused variable lookups inside the action to fail with "not found in service ''".

Fix: Value::Closure now stores the service_name from when it was created. When the closure is called, the body is evaluated with that saved service context instead of the caller's.

All existing tests pass.

### DIFF
--- a/meerkat-lib/src/runtime/ast/mod.rs
+++ b/meerkat-lib/src/runtime/ast/mod.rs
@@ -67,6 +67,7 @@ pub enum Value {
         params: Vec<String>,
         body: Box<Expr>,
         env: Vec<(String, Value)>,
+        service_name: String,
     },
     ActionClosure {
         stmts: Vec<ActionStmt>,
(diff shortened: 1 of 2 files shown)


JonathanAldrich (approved):
Looks good, thanks for the fix!

Issue #18, opened 2026-04-12
use Futures instead of polling for replies


JonathanAldrich:
Avoid polling for replies to a request; instead, the Manager should keep a list of pending reply Futures (or tokio channels to them) and when a reply comes in that matches the request_id, resolve the Future.

The architecture should allow specification or a timeout for replies and resolve the Future with an Error if no message comes before the timeout.

From @ShreejithBabuG's comment on PR 16: For the Future-based approach to replace the polling loops — the fix would be to store pending tokio::sync::oneshot channels in the Manager keyed by request_id, so incoming messages resolve the right Future directly without polling. This also fixes the message routing bug where a non-matching rid causes a response to be dropped.

PR #27, opened 2026-04-26
Implement #18: replace polling with oneshot channels for remote replies

nivethadhanakoti:
Replace the polling loops in remote_lookup and remote_action with tokio::sync::oneshot channels:

- Manager now holds a pending_replies: HashMap<u64, oneshot::Sender<MeerkatMessage>> keyed by request_id
- A new dispatch_network_events() method drains all incoming network events and routes each to the matching channel — this also fixes the message routing bug where a non-matching rid caused a response to be silently dropped
- Both remote_lookup and remote_action create a oneshot channel, register the sender, send the request, then await the receiver with a 15s timeout calling dispatch_network_events() each iteration
- Sleep interval reduced from 100ms to 10ms for better responsiveness

All 11 unit tests pass. All local integration tests pass.

### DIFF
--- a/meerkat-lib/src/runtime/ast/mod.rs
+++ b/meerkat-lib/src/runtime/ast/mod.rs
@@ -67,6 +67,7 @@ pub enum Value {
         params: Vec<String>,
         body: Box<Expr>,
         env: Vec<(String, Value)>,
+        service_name: String,
     },
     ActionClosure {
         stmts: Vec<ActionStmt>,
@@ -206,7 +207,7 @@ impl Display for Value {
             Value::Number { val } => write!(f, "{}", val),
             Value::Bool { val } => write!(f, "{}", val),
             Value::String { val } => write!(f, "\"{}\"", val),
-            Value::Closure { params, body, env } =>
+            Value::Closure { params, body, env, service_name } =>
                 write!(f, "fn({})[{:?}]{{{}}}", params.join(","), env, body),
             Value::ActionClosure { stmts, env, service_name } =>
                 write!(f, "action[{:?}][{}]{{{:?}}}", env, service_name, stmts),  
--- a/meerkat-lib/src/runtime/interpreter/evaluator.rs
+++ b/meerkat-lib/src/runtime/interpreter/evaluator.rs
@@ -237,7 +237,7 @@ mod tests {
         };
         let result = eval(&func_expr, &env, &mut ctx).await.unwrap();
         match result {
-            Value::Closure { params, body: _, env: captured_env } => {
+            Value::Closure { params, body: _, env: captured_env, .. } => {
                 assert_eq!(params.len(), 1);
                 assert_eq!(captured_env.len(), 1);
                 assert_eq!(captured_env[0].0, "a");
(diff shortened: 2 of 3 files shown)


JonathanAldrich (changes requested):
There's a lot of code duplication between the chunk of code starting at line 252 and the chunk starting at line 306.  Can you create a reusable abstraction that encapsulates this duplicated code, so that lines 252-282 get turned into just one function call (possibly with minimal additional bookkeeping)?

Also I think it may be possible to improve the architecture.  What if at the top level we add a `tokio::join!` call with two async functions passed to it.  One of the async functions is a message loop in manager.  The other runs the code for the Meerkat interpreter.  The two communicate via the oneshot channels.  Then if you send two messages you don't have two message loops competing with each other, which adds complexity and could result in inefficiency and possibly weird behavior. The message loop could, I think, be implemented without a timeout, and without using try; it just waits until some message is received, then dispatches it and waits for the next message.

I think you could also avoid a loop in the routines that send messages.  The timeout could be implemented with a `tokio::select!` that takes two async functions.  One results when a timeout timer expires, the other waits for a message.  Whichever resolves first, you proceed accordingly.

Can you think about these architectural changes and see what can be done?

nivethadhanakoti:
Hi Professor Aldrich,

I've pushed a follow-up commit addressing the first two points:

- Extracted the duplicated send + register + await pattern into a single
send_and_await_reply helper, so remote_lookup and remote_action each reduce
to one call to it plus a match on the result.
- Removed the loop entirely from send_and_await_reply — it now calls
dispatch_network_events() once, then uses a single tokio::select! with two
branches: one for the reply arriving via the oneshot channel, and one for
the 15s timeout. Whichever resolves first determines the result.

Regarding the tokio::join! architecture — I think it's the right long-term
direction. However, implementing it properly in Rust requires splitting
Manager's network I/O from its interpreter state, since both tasks would
need mutable access to different parts of Manager simultaneously.
Specifically, the background message loop would need to own the
NetworkActor and a reference to pending_replies, while the interpreter owns
the rest of Manager. This would likely involve an Arc<Mutex<...>> for
pending_replies and a channel for outbound messages. I'd be happy to tackle
this as a follow-up PR if that works for you.

JonathanAldrich:
Ok so the code looks good now!  But, I think there is a problem somewhere, when I test it using the simple distributed test case in the top level project README (basically running s1.mkt on one server and test_client.mkt on another machine), the test_client process always times out.  I am not able to get test_client to pass, but it does pass with the original PR.  Can you check it out and see what might be causing this?

JonathanAldrich:
By the way I agree that we should separate out the tokio::join! architecture change into a separate issue.  I think it may be possible to do this without Arc<Mutex<...>>.  But it might require an actor messaging approach.  Will think about it and post an issue with more details, and we can discuss further.

nivethadhanakoti:
Found and fixed the issue. When we removed the loop from
send_and_await_reply, dispatch_network_events() was only called once before
the select!. If the reply from the remote server hadn't arrived at that
point, nobody would call it again — so the oneshot channel would never get
fulfilled and the request would always time out.

The fix restores the loop with a pinned tokio::select! timeout, so
dispatch_network_events() is called each iteration until the reply arrives
or the timeout fires. The key improvements over the original code are still
in place: the duplication is gone (single send_and_await_reply helper),
dispatch_network_events routes all incoming events to the right channel
fixing the message routing bug, the timeout uses a pinned future so it
isn't reset each iteration, and the sleep interval is reduced from 100ms to
10ms.

The no-loop version will be possible once the tokio::join! background
message loop is implemented as a follow-up.

JonathanAldrich (approved):
Thanks, it works now!

PR #33, opened 2026-05-24
Issue 32 server remote import

nivethadhanakoti:
**Root cause:** `create_service` was called before the network actor was started and before remote services were registered. So any remote lookup during service initialization failed immediately with `Variable 'y' not found in service 's1'`.

**Fix:** Reordered `run_server` so the network is started, remote services are registered, and the network is wired into the manager — all before `create_service` is called.

Also added a `-p`/`--port` flag (default 9000) to allow running multiple servers on different ports when testing on the same machine.

**Tested:**
- `cargo run -- -s -f meerkat/tests/s1.mkt -p 9000`
- `cargo run -- -s -f meerkat/tests/test_client.mkt -p 9001 -i <s1_url>` → `Service 's2' loaded` 
- `cargo run -- -f meerkat/tests/test_client.mkt -i <s1_url>` → `@test(s2) passed`


JonathanAldrich (approved):
Looks good, thanks for the fix!

Issue #40, opened 2026-05-31
fix bug in transitive action execution


JonathanAldrich:
To reproduce:

1. Run `s1.mkt` in server mode (`-s`)
2. Run `test_client.mkt` also in server mode, importing `s1` from the first process
3. Run `s3.mkt` importing `s2` from the 2nd process

I get the error message
```
Error: "Service error: Lookup error: Variable 'inc_x' not found in service 's1'"
```

dustinjuliano:
I am working on this and should have a draft PR soon.

dustinjuliano:
Update from 2026-06-01 meeting with Professor Aldrich:

- We are converging on the idea of resolving this issue using [Execution Location](https://github.com/meerkat-lang/meerkat/wiki/Requirements#execution-location) from the Wiki, which would eliminate the proxy closure idea from now closed PR #41 and introduce an RPC style dispatch/execution.
- Instructed to leave this status open for now and to return to static type system (Issue #34).
- Historical: Working on this issue resulted in the network orchestrator proposal PR #42 , which was split off from PR #41.

PR #37, opened 2026-05-29
[abandoned for stacked PRs] feat(types): static bi-directional type system

dustinjuliano:
Introduces bi-directional type checking system. Proposed implementation for Issue #34. This is currently under re-organization and is likely to become part of a Stacked PR soon.

PR #42, opened 2026-06-01
feat(tooling): network orchestrator

dustinjuliano:
This PR is a feature proposal for a network orchestrator. As instructed, I isolated this from PR #41 for a clean merge after closing it.

This orchestrator was originally introduced to automate replication for automating build and test of Issue #40.

## Summary by CodeRabbit

* **New Features**
  * Added a --local CLI flag to bind network listeners to loopback for safe local development
  * Added a manifest-driven network orchestration CLI with automatic service discovery, relay/gateway support, and state dumping

* **Tests**
  * Large suite of new integration and unit test fixtures covering multi-node topologies, gateway/relay scenarios, invalid-config cases, and a mock client harness

* **Chores**
  * Updated ignore rules to exclude temporary folders


scripts/default_manifest.txt
avishj, line 1: It is debatable but I think using a different file extension other just `.txt` might be useful if we plan to have several manifests for different e2e tests so that it would be obvious while browsing the codebase that those files represent a manifest for a network of meerkat nodes. The content could continue to remain plaintext with the same format as currently.

scripts/default_manifest.txt
```diff
@@ -0,0 +1,18 @@
+# Meerkat Network Orchestration Manifest
+# Format: node_name:file_path:port:imports
```
avishj: It might also be useful to create a simple helper somewhere that enforces a schema for this format, especially if we plan to test more complicated setups with this down the line. I was initially thinking of something like jsonschema but I don't think that works well here since it wouldn't capture semantics. A simple python function that validates structure + semantics (duplicate nodes/ports or client having no imports or invalid imports, etc.) might be useful.

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+
+                if os.path.exists(log_file_path):
+                    with open(log_file_path, "r") as lf:
+                        content = lf.read()
+                        matches = re.findall(r"Service URL:\s+(\S+)", content)
+                        if matches:
+                            for url in matches:
+                                svc_name = url.split('/')[-1]
+                                service_urls[svc_name] = url
+                            svc_url = matches[0]
```
avishj: Breaking on the first service URL might be an issue, since we can have multiple services in a node. If within the 0.2s, one service URL gets printed and the other doesn't, we can miss one completely. We would perhaps need something like a services ready log or similar. Something after this: https://github.com/meerkat-lang/meerkat/blob/21909664e4769e202da9628b686e3b734928fdf8/meerkat/src/main.rs#L101-L106

like a `println!("Services ready!");`, which is then checked for here before setting `url_found = True`.

Admittedly this is fairly low probability to occur IRL but can cause hard to debug transient bugs.

scripts/default_manifest.txt
```diff
@@ -0,0 +1,18 @@
...
+s1: meerkat/tests/s1.mkt: 9001:
+
+# 2. Intermediate gateway server, importing s1
+s2: meerkat/tests/test_client.mkt: 9002: s1
+
+# 3. Downstream consumer/test runner, importing s2
+s3: meerkat/tests/s3.mkt: client: s2
+
+# 4. Multi-service backend server (regression test for HashMap non-determinism)
+s_multi: meerkat/tests/s_multi.mkt: 9003:
```
avishj: Is this file `meerkat/tests/s_multi.mkt` supposed to be included with this PR? Or it is just an example?

scripts/default_manifest.txt
```diff
@@ -0,0 +1,18 @@
...
+
+# 5. Client for multi-service backend server testing
+test_multi: meerkat/tests/test_multi_client.mkt: client: s_multi_1, s_multi_2
```
avishj: Same comment as R15.

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+            for imp in [i.strip() for i in imports.split(',') if i.strip()]:
+                resolved_url = service_urls.get(imp)
+                if not resolved_url:
+                    url_file = os.path.join(LOG_DIR, f"{imp}.url")
+                    if os.path.exists(url_file):
+                        with open(url_file, 'r') as uf:
+                            resolved_url = uf.read().strip()
+                
+                if not resolved_url:
+                    print(f"Error: Node '{node_name}' imports '{imp}', but '{imp}' has not been started yet.")
```
avishj: Is this supposed to be Node or Service? Since `imp` is checked against `service_urls`?

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+                    print(lf.read())
+                cleanup(exit_code=1)
+
+            log_file.close()
+            service_urls[node_name] = svc_url
+            
+            # Save URL file for team integration
+            url_file_path = os.path.join(LOG_DIR, f"{node_name}.url")
+            with open(url_file_path, "w") as uf:
+                uf.write(svc_url)
```
avishj: Same as R177, shouldn't it account for multiple services in a Node?

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+                if res.returncode != 0:
+                    print(f"[{node_name}] Execution failed with code {res.returncode}. Output above.")
+                    sys.exit(res.returncode)
+            except Exception as e:
+                print(f"[{node_name}] Execution failed: {e}")
+                sys.exit(1)
+        else:
+            # Server Node (runs in background)
+            print(f"[{node_name}] Starting server node on port {port} running '{file_path}'...")
+            cmd = ["./target/debug/meerkat", "-s", "-f", file_path, "-p", port] + import_flags
```
avishj: This assumes we are on workspace root and that we want to run the debug binary...maybe we should use cargo run instead? Or at least document the requirement to run it on workspace root. Release binary support is probably not that important, but might be a nice to have a flag.

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+
+        if port.lower() == "client":
+            # Client Node (runs in foreground)
+            print(f"[{node_name}] Starting client node running '{file_path}'...")
+            cmd = ["./target/debug/meerkat", "-f", file_path] + import_flags
+            print(f"Executing: {' '.join(cmd)}")
+            print("---------------------------------------------------")
+            
+            try:
+                res = subprocess.run(cmd)
```
avishj: We should perhaps consider some default timeout to prevent a hang.

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+import sys
+import time
+import subprocess
+import signal
+import re
+import shutil
+import atexit
+
+# Enable loopback-only local bindings by default
+os.environ["MEERKAT_LOCAL"] = "1"
```
avishj: I couldn't find this env being used in the lib. Is this being added somewhere else?

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+signal.signal(signal.SIGINT, cleanup)
+signal.signal(signal.SIGTERM, cleanup)
+
+def main():
+    global processes
+
+    # Parse arguments
+    manifest_path = "scripts/default_manifest.txt"
+    if len(sys.argv) > 1:
+        if sys.argv[1] in ("-h", "--help"):
```
avishj: No args should ideally also trigger help message.

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+
+            if not url_found:
+                print(f"Error: Timeout waiting for server '{node_name}' to start. Log output:")
+                log_file.close()
+                with open(log_file_path, "r") as lf:
+                    print(lf.read())
+                cleanup(exit_code=1)
+
+            log_file.close()
+            service_urls[node_name] = svc_url
```
avishj: I'm confused, why are we mapping node to the first matching `svc_url`?

scripts/run_network.py
```diff
@@ -0,0 +1,205 @@
...
+            
+            # Save URL file for team integration
+            url_file_path = os.path.join(LOG_DIR, f"{node_name}.url")
+            with open(url_file_path, "w") as uf:
+                uf.write(svc_url)
+                
+            print(f"[{node_name}] Started successfully! Service URL: {svc_url}\n")
```
avishj: Same concern with multiple services on a Node.

avishj (changes requested):
Hey @dustinjuliano, this protocol is a great idea, I had some questions/suggestions which I put down as comments. Could we have a huddle to help me understand or you could respond via comments, whichever you prefer?

dustinjuliano:
> Hey @dustinjuliano, this protocol is a great idea, I had some questions/suggestions which I put down as comments. Could we have a huddle to help me understand or you could respond to comments, whichever you prefer?

Thank you! I will respond to the individual comments and propose fixes where I can and will start working on this ASAP. Just let me know on Slack if you still want to meet after and I'll be there.

avishj:
> > Hey @dustinjuliano, this protocol is a great idea, I had some questions/suggestions which I put down as comments. Could we have a huddle to help me understand or you could respond to comments, whichever you prefer?
> 
> Thank you! I will respond to the individual comments and propose fixes where I can and will start working on this ASAP. Just let me know on Slack if you still want to meet after and I'll be there.

That sounds great, I'll await your updates. We can coordinate via Slack post that as needed. Thanks!

scripts/default_manifest.txt
dustinjuliano: Addressed in commit 312a73c. I agree 100% and I propose `.mkn`. Kept as plaintext for content. In the presentation I mentioned possibly going for JSON, which may then go better with your idea for JSON Schema. Will discuss in other comments or perhaps in person.

scripts/default_manifest.txt
dustinjuliano: Addressed in commit 312a73c. Good catch. I rewrote the default manifest to be suitable just for this PR; it now runs correctly because the old manifest was reproducing an issue/bug. This new manifest removes reference to that, includes multi-service Meerkat programs to run it, and demonstrates the correct execution of the network orchestrator.

scripts/default_manifest.txt
dustinjuliano: Addressed in commit 312a73c.

scripts/run_network.py
dustinjuliano: Addressed in commit 312a73c. 100% agree. The script now uses `cargo run` everywhere it would have previously called the binary. This is vastly superior and avoids the earlier warning I put in the description for this PR, which I will now delete.

scripts/run_network.py
dustinjuliano: Addressed in commit 312a73c. The slice out from the PR to this PR missed the local binding logic for the Meerkat library. However, upon reflection, I realized the environment variable would add more friction for team members than a CLI switch, so I propose `--local` and added the logic for client and servers to bind to local host, which has numerous benefits for offline, airplane mode, and cybersecurity while testing our network protocols and systems. Very good catch and apologies for missing it, but it now leads to what I think is a better design (the CLI).

dustinjuliano:
Excellent catches, @avishj. I submitted commit 312a73c and am now moving on to work on the other comments as one or more additional commits. Will have them addressed as soon as possible.

scripts/default_manifest.txt
avishj: Sounds good!

avishj:
> Excellent catches, @avishj. I submitted commit [312a73c](https://github.com/meerkat-lang/meerkat/commit/312a73cc6a359cfca390bfdc95f40382f2adcc40) and am now moving on to work on the other comments as one or more additional commits. Will have them addressed as soon as possible.

Sure, let me know once you're done and I'll take a look again.

avishj:
@coderabbitai review

meerkat/src/repl.rs
```diff
@@ -56,7 +56,8 @@ pub async fn run_repl(
     if !remote_url_map.is_empty() {
         let mut n = meerkat_lib::net::NetworkActor::new(meerkat_lib::net::types::NodeType::Server).await
             .map_err(|e| format!("Network error: {}", e))?;
-        let listen_addr = meerkat_lib::net::Address::new("/ip4/0.0.0.0/tcp/0");
+        let listen_ip = if manager.local { "127.0.0.1" } else { "0.0.0.0" };
+        let listen_addr = meerkat_lib::net::Address::new(&format!("/ip4/{}/tcp/0", listen_ip));
         n.handle_command(meerkat_lib::net::NetworkCommand::Listen { addr: listen_addr }).await;
```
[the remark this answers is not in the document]
dustinjuliano: This does not seem related to the feature of the Network Orchestrator. The affected lines were added to implement local loopback. It may be better brought up as a GitHub Issue?

meerkat/src/main.rs
```diff
@@ -179,15 +192,18 @@ async fn run_client(
...
 
     // Start network if we have remote imports
     let mut net: Option<NetworkActor> = None;
     if !remote_url_map.is_empty() {
         let mut n = NetworkActor::new(NodeType::Server).await
             .map_err(|e| format!("Network error: {}", e))?;
-        let listen_addr = Address::new("/ip4/0.0.0.0/tcp/0");
+        let listen_ip = if local { "127.0.0.1" } else { "0.0.0.0" };
+        let listen_addr = Address::new(&format!("/ip4/{}/tcp/0", listen_ip));
         n.handle_command(NetworkCommand::Listen { addr: listen_addr }).await;
```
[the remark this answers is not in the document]
dustinjuliano: I am not sure this problem was introduced by adding local loopback. I will have to compare the original with the affected lines. If it was not introduced by having a local loopback feature then this is an improvement that is outside the scope of the network orchestrator and I would have to ask if we would open an issue or patch it in here; it risks expanding the scope of this PR.

meerkat/src/repl.rs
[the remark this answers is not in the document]
dustinjuliano: Only if you have been granted permission to do so. If you do so, then we can mark this sub-thread as resolved.

meerkat/src/main.rs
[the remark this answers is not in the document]
dustinjuliano: Only if you have been given permission. In the comment after this one, you raise a similar concern; if these two issues can be fused into one then that is more logical. We can then mark this sub-thread as resolved.

scripts/run_network.py
```diff
@@ -0,0 +1,204 @@
...
+    for node_name, proc in processes:
+        if proc.poll() is None:
+            print(f"Stopping server '{node_name}' (PID: {proc.pid})...")
+            try:
+                proc.terminate()
+                proc.wait(timeout=2)
+            except subprocess.TimeoutExpired:
+                proc.kill()
+            except Exception:
+                pass
```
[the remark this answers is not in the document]
dustinjuliano: This looks to be in scope of the feature proposal and I will investigate as soon as possible.

dustinjuliano:
@avishj Status update: Thursday was heavy meetings for REUSE so I did not get a chance to address the remaining change requests; goal is to do that this morning. I also meant to say before that I am always open to meeting and we can use Slack DMs to exchange schedules.

scripts/run_network.py
dustinjuliano: Addressed in commit 80212ea.

scripts/run_network.py
dustinjuliano: Fixed by commit 3a78b0d.

dustinjuliano:
For the remaining comments I am preparing a major feature upgrade to the orchestrator. In my original presentation under future considerations I mentioned JSON for configuration/manifest. This new design will do that. I grouped this design change with the comments because it addresses the race condition, multi-service, gateways for clients hosting services, parsing, validation/semantics/schema, and also synchronization with the Rust side. Commit coming soon.

scripts/default_manifest.txt
dustinjuliano: Addressed in commit 518c4b0. It is now based on JSON. The `Manfiest` class entails its semantics and validation before being passed on for use by the system. It must be correct during its construction or it raises an exception. No external dependencies for the schema; pure Python error-checking.

scripts/run_network.py
dustinjuliano: Addressed in commit 518c4b0. This was one of the largest issues to solve and was an extremely important finding. Will explain in larger comment at end.

scripts/run_network.py
dustinjuliano: Addressed in commit 518c4b0; completely new PeerID based discovery mechanism. See larger commit comment.

scripts/run_network.py
dustinjuliano: Addressed in commit 518c4b0.

scripts/run_network.py
```diff
@@ -0,0 +1,332 @@
...
+                    exit_orchestrator(1, processes, f"Import resolution error for node '{node_name}'")
+
+                import_flags.extend(["-i", resolved_url])
+
+        if port.lower() == "client":
+            log_file_path = os.path.join(clients_log_dir, f"{node_name}.log")
+            # Client Node (runs in background)
+            print(f"[{node_name}] Starting client node running '{file_path}'...")
+            cmd = ["cargo", "run", "-p", "meerkat", "--", "--local", "-f", file_path] + import_flags
+            print(f"Executing: {' '.join(cmd)}")
```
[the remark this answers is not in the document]
dustinjuliano: Addressed in commit 518c4b0. Please hold review until requested by one of the human team members as I still need to communicate changes.

dustinjuliano:
Explanation of commit 518c4b0. Major rewrite of the Meerkat Network Orchestrator (MKN). This is an object-oriented design with clean separation of concerns, a dedicated test script and test suite. It now uses JSON for the manifest format and is fully validated in the `Manifest` class. The class design models the topology and design of the Rust side of Meerkat, with `Node`s and `Services` and a main `NetworkOrchestrator` class to hold the state of the program.

To future proof the design the way in which name spaces and relative naming is done had to be modeled by MKN. This is based on meetings with Professor Aldrich who informed me about the subtleties of gateways and clients being able to offer services through relay. To avoid merge conflicts, I ensured no changes to the Rust side; test mocks were set up and the logic is fully present in MKN to model in the manifest and launch in the right order the relays necessary for clients to offer services through their gateway servers, and all the naming is now hierarchical and split into three separate spaces:

- Services local to the current node
- Services remote to the current node
- Services relayed by the current node

This reflects the `libp2p` framework relay URI/URL scheme of being able to refer to a client that is behind a server and it cleanly can distinguish names of services as separate namespaces from the services that might be local to the relay itself without clashing. This may even be a suggestion for now name resolution could be done on the Rust side. Will discuss with Professor Aldrich next meeting (which is today in a few moments).

Another important change is that MKN intelligently can discover services correctly now in arbitrary order and will know when all services are loaded or not by the message sent from the Rust side that the node is now fully online.

Misc other things such as ensuring docstrings and good form throughout. Constants moved into manifest for better maintenance. I believe it may be ready for further review and is substantially improved from the original quick script.

scripts/mkn.py
```diff
@@ -0,0 +1,711 @@
...
+            if "cmd" in node_def:
+                if not isinstance(node_def["cmd"], list) or not all(isinstance(x, str) for x in node_def["cmd"]):
+                    raise ValueError(f"Node '{alias}' 'cmd' must be a list of strings")
+                    
+            # port check
+            if "port" in node_def:
+                if node_type == "client":
+                    raise ValueError(f"Client node '{alias}' cannot specify a port number")
+                if node_def["port"] is not None and not isinstance(node_def["port"], int):
+                    raise ValueError(f"Server node '{alias}' 'port' must be an integer")
```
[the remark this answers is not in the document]
avishj: Agreed, this could be improved with a concrete check.

scripts/mkn.py
```diff
@@ -0,0 +1,711 @@
...
+        """
+        for node in self.nodes_by_alias.values():
+            if node.is_started:
+                continue
+                
+            imports_resolved = True
+            if node.gateway:
+                dep_node = self.nodes_by_alias[node.gateway]
+                if not dep_node.is_online:
+                    imports_resolved = False
```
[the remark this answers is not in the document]
avishj: This is an edge case but possible in a race situation. You could add a fix by making sure `dep_node.peer_id` is truthy.

scripts/mkn.py
```diff
@@ -0,0 +1,711 @@
...
+                raise ValueError(f"Node '{alias}' type must be 'server' or 'client', got '{node_type}'")
+                
+            # file / cmd check
+            if "file" not in node_def and "cmd" not in node_def:
+                raise ValueError(f"Node '{alias}' must specify either 'file' or 'cmd'")
+            if "file" in node_def and not isinstance(node_def["file"], str):
+                raise ValueError(f"Node '{alias}' 'file' must be a string")
+            if "cmd" in node_def:
+                if not isinstance(node_def["cmd"], list) or not all(isinstance(x, str) for x in node_def["cmd"]):
+                    raise ValueError(f"Node '{alias}' 'cmd' must be a list of strings")
```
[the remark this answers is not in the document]
avishj: Valid issue, should be fixed or at least print an info/error.

scripts/test_mkn.py
```diff
@@ -0,0 +1,149 @@
+import subprocess
+import json
+import sys
+import re
+
+def run_cmd(args):
+    proc = subprocess.run(args, capture_output=True, text=True)
+    return proc.returncode, proc.stdout, proc.stderr
```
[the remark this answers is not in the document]
avishj: Valid concern.

scripts/test_mkn.py
```diff
@@ -0,0 +1,149 @@
...
+        
+        if code == 0:
+            print(f"FAIL: Expected non-zero code for {name}")
+            print(f"Actual output: {stdout.strip()}")
+            return False
+            
+        if expected_error not in stdout:
+            print(f"FAIL: Expected validation error regarding '{expected_error}'")
+            print(f"Actual output: {stdout.strip()}")
+            return False
```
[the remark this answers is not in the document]
avishj: Valid concern.

scripts/mkn.py
```diff
@@ -0,0 +1,711 @@
...
+                else:
+                    print(f"[{node.alias}] Client node completed with exit code {status}.")
+                    self.dump_node_log(node)
+                    
+            # B. Check for startup timeout
+            elif not node.is_online:
+                if now - node.start_time > self.startup_timeout:
+                    print(f"\nError: Timeout waiting for node '{node.alias}' to initialize after {self.startup_timeout} seconds.")
+                    self.dump_node_log(node)
+                    raise TimeoutError(f"Timeout waiting for node '{node.alias}' to start")
```
avishj, line 564: Wouldn't clients always end up going here? Since you define at L420-21 the `is_online` based on the server print message. So `not node.is_online` will always be true for clients then? Which would then make branch `C` impossible to reach.

scripts/mkn.py
```diff
@@ -0,0 +1,711 @@
...
+        
+        Args:
+            manifest: A validated Manifest configuration instance.
+        """
+        self.manifest = manifest
+        self.nodes_by_alias = {} # Dict[alias, Node]
+        self.nodes_by_peer_id = {} # Dict[peer_id, Node]
+        
+        # Session config
+        self.session_id = datetime.now().strftime("%Y%m%d_%H%M%S")
```
avishj: Use PID or UUID or something similar, %S isn't likely to be unique in batch tests.

scripts/mkn.py
```diff
@@ -0,0 +1,711 @@
...
+            # gateway check
+            if "gateway" in node_def:
+                gateway = node_def["gateway"]
+                if gateway not in self.nodes_by_alias:
+                    raise ValueError(f"Node '{alias}' specifies gateway '{gateway}' which does not exist in the manifest")
+                    
+            # imports check
+            for imp in node_def.get("imports", []):
+                dep_alias = imp.split('.')[0]
+                if dep_alias not in self.nodes_by_alias:
```
avishj, line 232: We check `dep_alias` but we don't check services on a given node. An incorrect service would hang at `spawn_resolved_nodes()` right?

scripts/test_mkn.py
```diff
@@ -0,0 +1,149 @@
+import subprocess
+import json
+import sys
+import re
+
+def run_cmd(args):
+    proc = subprocess.run(args, capture_output=True, text=True)
```
avishj: We would need some way to ensure meerkat processes are killed else we will have orphans which can impact later tests in a batch. I'm not familiar with all the ways to do this but apparently a process group might work.

avishj (changes requested):
Great improvements, a few items could be improved, which I have highlighted in comments. Let me know if you need clarification on any.

avishj:
Also, can we add `scripts/__pycache__` to `.gitignore`? Running `mkn` generates temp files there. Perhaps the `mkn` could also cleanup after itself?

avishj:
Same for `/tmp/logs/mkn/**` but perhaps this should be gated behind a flag so that runs with errors don't get their logs deleted.

scripts/mkn.py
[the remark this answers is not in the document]
dustinjuliano: Corrected in 37ac85f.

scripts/mkn.py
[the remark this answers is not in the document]
dustinjuliano: Corrected by d47fe60.

scripts/mkn.py
dustinjuliano: Corrected in cde9474.

scripts/mkn.py
dustinjuliano: Addressed in commit 2178d0f. Required moderate changes to fix. Summary:

- Two timeouts exist: startup and running
- Previously: servers had startup, but not running timeout. Clients had both startup and running timeout. However, clients do not print the expected stdout message to signal to MKN to change state from startup timeout to running timeout; result is clients could be terminated wrongly for running longer than the startup timer
- Correction: Now use the dependency information in the manifest to (very) efficiently compute whether nodes are dependent or not. This future proofs the design so that if and when relays are added, MKN will intelligently apply startup timers or not
- Added tests for this logic

scripts/test_mkn.py
dustinjuliano: Resolved in 6c76258, a substantial change to the test driver; enhanced proposal by making it cross-platform.

dustinjuliano:
@avishj , I believe I have now covered all new review comments. Professor @JonathanAldrich on Friday wanted me to transition to #37 as primary focus; however, I did not want to leave this PR in an unsteady state. There is also #55 coming soon for the one-time format update. 

I believe this PR is in a much healthier state thanks to your insightful reviews and I would like to propose a freeze until the decision point of #55. If we do rebase or merge all PRs, I believe this PR is in a good state for either choice. I am open to direction of course and this is only a proposal so I can focus on #37, which I will turn my attention to fully today as well.

avishj (approved):
LGTM!

JonathanAldrich:
Looks good to me too.  Thanks @dustinjuliano for the PR and @avishj for the reviews.

Issue #44, opened 2026-06-03
Make transactions work across distributed services


JonathanAldrich:
It should be possible to execute a transaction that starts with an action defined in service `s1`, where that action then composes another action defined in some other service `s2`.  It should work even if `s1` and `s2` are on different network nodes.  The original action might look something like this:

```
action {
   x = x + 1;
   do s2.update;
}
```

where `update` is an action defined in `s2` which modifies variables in `s2`.  Of course this should work transitively as well (e.g. if `s2.update` composes a third action from service `s3`).

Design suggestions:
 * All transaction data structures that mention variables should be revised to use a pair of (service identity, variable).
 * In the pair, the service should not be merely a name, as that name might not be bound depending on what scope the transaction is originally executed in.  Furthermore, as we get multiple clients executing their own local service that has the same name as other clients, names become ambiguous. So the service should be a _service identity_.  The easiest and most practical way to represent a service identity is simply the global network address of the service (i.e. the address reachable from a public internet node, which possibly includes a websocket relay to a web client).

Issue #24, opened 2026-04-25
asynchronous updates


JonathanAldrich:
The new `watch` functionality is great!  However, it is implemented essentially with polling.  That's OK for testing purposes, but we need the architecture to support asynchronous updates (i.e. change notifications).  That way someone looking at an app can see the data in the app updating in real time without taking any action or requiring constant polling messages.

Consider program `s1.mkt` and `test_client.mkt`.  When the def `z` in `s2` is initialized, it depends on def `y` in `s1`.  Therefore, it will register itself as a listener on `s1.y`.  `s1.y` will keep track of all its listeners.  Whenever `s1.y` changes, it sends a change message to each of its listeners.  The change messages are managed by the manager, which can notify local services as well as remote services.  When `s2` gets the change message, it re-evaluates `z` and stores the new value.  For efficiency, if `z` depends on several defs/vars from other services, it should cache their values so it does not need to do any additional round-trip communication when it gets a change message.

To demonstrate that this is working, and help with future debugging, add a command-line flag to print change notifications asynchronously as they arrive (`watch`-style, but not waiting for a REPL command to be entered, and done for all defs and vars that are depended on by something).

Optional: `watch` itself could be reimplemented to use change notifications.  But it is also OK to leave it with polling for now.

JonathanAldrich:
Andrew Zhou will work on this

JonathanAldrich:
The implementation of this should replace Manager::propagate

PR #52, opened 2026-06-06
Implements #44: a transaction starting in one service can compose actions in other services, atomically, including across network nodes and transitively.

nivethadhanakoti:
Approach (Option B, per our discussion): the originating node ships a composed action to its owning node along with the shared transaction id. The owner executes under that id, holds its locks and buffers its writes, and waits for a Commit or Abort from the originator. Commit and Abort forward down the chain, so transitive composition (s1 to s2 to s3) commits or rolls back as one unit.

Stage 1 (local cross-service):
- New ServiceId type; every service carries one.
- All transaction state (locked, read_cache, written) is keyed by (ServiceId, variable) so the same name in different services never collides.
- The transaction threads through nested cross-service do, so an action in s1 can compose s2's action on one node.

Stage 2 (cross-node):
- TxnId is now serializable and globally unique (node id plus timestamp).
- New Commit, CommitResponse, Abort, AbortResponse messages; ActionRequest carries an optional shared txn id.
- Participant nodes execute composed actions under the shared id and hold them until Commit or Abort. Commit and Abort forward to sub-participants for transitive chains.

Testing:
- Unit: cross-service composition, read-then-write upgrade, nested do, no partial writes on failure (17 lib tests pass).
- Cross-node (.mkt): two-node commit, abort releasing remote locks, and a three-node transitive chain (s1 to s2 to s3) verified to commit all writes.

Limitations, noted for transparency:
- ServiceId falls back to the bare name for local-only execution with no network address (unambiguous on a single node); address-based identity is used whenever networking is present.
- Remote reads within a transaction are not lock-protected (Option B ships actions for remote writes; per-variable remote read locking would be the fuller Historiographer protocol). 2PC uses execute-and-hold as an implicit prepare rather than a separate prepare-vote round.

## Summary by CodeRabbit

* **New Features**
  * Distributed transactions across services: prepare/commit/abort coordination, stable service identities, and transaction-aware remote reads/actions.

* **Tests**
  * Added and expanded end-to-end tests for single-node, two-node, and three-node cross-service transaction scenarios.

* **Documentation**
  * New guidance and runnable examples documenting distributed transaction behaviors and test coverage.

* **Chores**
  * Miscellaneous typo and formatting fixes across the codebase.


meerkat-lib/src/runtime/txn.rs
```diff
@@ -4,33 +4,47 @@
...
-            timestamp: SystemTime::now(),
-            iteration: 0,
-        }
+    /// Create a new transaction id originating from the given node.
+    pub fn new(node_id: u64) -> Self {
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.as_nanos())
+            .unwrap_or(0);
+        TxnId { timestamp, node_id, iteration: 0 }
```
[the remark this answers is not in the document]
JonathanAldrich: Yes, we should ensure transaction uniqueness.  CodeRabbit's suggestion would work; we could also cache the last nano-time used and bump it by one if the time has not changed.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -490,40 +564,159 @@ impl Manager {
...
+                Err(e) => { exec_error = Some(e); break; }
+            }
+        }
+        if let Some(e) = exec_error {
+            // Execution failed: release any locks acquired; nothing to hold.
+            self.release_locks(&txn.locked, &txn.id);
+            return Err(e);
+        }
+        // Prepared: hold locks and buffered writes until commit/abort.
+        self.pending_txns.insert(tid, txn);
```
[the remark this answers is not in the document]
JonathanAldrich: Yes, this needs to be fixed.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -490,40 +564,159 @@ impl Manager {
...
+    /// Originator side: tell a participant to abort, awaiting acknowledgement
+    /// so its locks are released before we return (and the process may exit).
+    async fn send_abort(&mut self, addr: Address, tid: &TxnId) {
+        use std::sync::atomic::{AtomicU64, Ordering};
+        static NEXT_ABORT_ID: AtomicU64 = AtomicU64::new(1);
+        let request_id = NEXT_ABORT_ID.fetch_add(1, Ordering::SeqCst);
+        let reply_to = self.local_reply_addr().await;
+        let msg = MeerkatMessage::Abort { request_id, txn_id: tid.clone(), reply_to };
+        // Best-effort: if the ack times out we still proceed (locks will be
+        // reclaimed by the participant when it next reconciles).
```
JonathanAldrich: "locks will be reclaimed by the participant when it next reconciles" - do we actually do this?

meerkat/src/main.rs
```diff
@@ -165,6 +170,29 @@ async fn run_server(prog: Vec<Stmt>, remote_url_map: std::collections::HashMap<S
...
+                        }
+                        MeerkatMessage::Abort { request_id, txn_id, reply_to } => {
+                            manager.abort_participant(&txn_id).await;
+                            if let Some(net) = manager.network.as_mut() {
+                                net.handle_command(NetworkCommand::SendMessage {
+                                    addr: Address::new(&reply_to),
+                                    msg: MeerkatMessage::AbortResponse { request_id },
+                                }).await;
+                            }
+                        }
```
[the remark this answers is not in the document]
JonathanAldrich: I think the code for commit is fine (the transaction is already committed; if the forwarding fails, there's not much we can do except report a problem, which the code does).  We should probably indeed add retries but let's leave that for later.  Question: do we want/need to report that the abort failed?

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -33,6 +42,20 @@ impl Manager {
...
+            node_id: Self::random_node_id(),
+            pending_txns: HashMap::new(),
+        }
+    }
+
+    /// Compute the global identity of a service owned by this node. When the
+    /// node has a network address, the identity is that address plus the service
+    /// slug; otherwise it falls back to the bare name for local-only execution.
+    async fn service_identity(&mut self, name: &str) -> ServiceId {
+        let addr = self.local_reply_addr().await;
```
JonathanAldrich: This is never going to initially reply with an empty string and then later with an actual address, is it?  because we would want this to be stable.

meerkat-lib/src/runtime/interpreter/executor.rs
```diff
@@ -32,10 +32,10 @@ pub async fn execute(
...
                 Value::ActionClosure { stmts, env: closure_env, service_name: action_svc } => {
                     if manager.remote_services.contains_key(&action_svc) {
-                        // Remote actions don't yet participate in the local
-                        // transaction; distributed lock coordination is future work.
-                        manager.remote_action(&action_svc, stmts, closure_env).await?;
-                        tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+                        // Ship the action to its owning node under the shared
+                        // transaction (Option B). The remote node executes and
+                        // holds; our commit/abort decides its fate.
+                        manager.remote_action(&action_svc, stmts, closure_env, txn.as_deref_mut()).await?;
```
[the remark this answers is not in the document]
JonathanAldrich: I think this is correct, we should register the remote participant before sending the message.  A failure should lead to an abort, and then the abort will notify the remote participant so it can release locks (if it grabbed any).

If there's no path where the remote participant could fail and retain locks, then this would be unnecessary.  I'm guessing at least timeouts could lead to this problem though.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -395,6 +450,25 @@ impl Manager {
         }
     }
 
+    /// Resolve an in-scope service name to its global ServiceId. For a known
+    /// local service this is its stored id; otherwise it falls back to a
+    /// name-based id (used until remote service identities are wired up).
```
JonathanAldrich: "used until remote service identities are wired up" - can this create inconsistencies that lead to bugs?

meerkat/src/main.rs
```diff
@@ -151,8 +151,13 @@ async fn run_server(prog: Vec<Stmt>, remote_url_map: std::collections::HashMap<S
...
                         }
-                        MeerkatMessage::ActionRequest { request_id, service, stmts, env: action_env, reply_to } => {
-                            let result = manager.execute_action_with_env(&service, &stmts, &action_env).await;
+                        MeerkatMessage::ActionRequest { request_id, service, stmts, env: action_env, reply_to, txn_id } => {
+                            // Part of a distributed transaction: execute under the
+                            // shared id and hold. Standalone: commit immediately.
+                            let result = match txn_id {
+                                Some(tid) => manager.execute_action_participant(&service, &stmts, &action_env, tid).await,
+                                None => manager.execute_action_with_env(&service, &stmts, &action_env).await,
+                            };
```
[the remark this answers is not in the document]
JonathanAldrich: (yes, this should be fixed - details in execute_action_participant)

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -490,40 +564,159 @@ impl Manager {
...
         match exec_error.or(commit_error) {
             Some(e) => Err(e),
             None => Ok(()),
         }
     }
 
+    /// Apply a transaction's buffered writes to the owning services, record the
+    /// writing transaction, and propagate to dependent defs. Shared by local
+    /// commit and by a participant committing on a remote Commit message.
+    async fn apply_committed_writes(&mut self, txn: &Transaction) -> Result<(), EvalError> {
```
JonathanAldrich: If this errors out, the caller tries to abort, but it's too late; the writes have been executed.

So this should not return a result; it should always succeed.  Likewise, propagate should not return a result; it should always succeed.  The only possible failure is inside propagate, and it should lead to adding the propagation to a data structure that will enable retries.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -490,40 +564,159 @@ impl Manager {
...
                     commit_error = Some(e);
-                    break;
                 }
             }
+        } else {
+            // Roll back: tell every participant to abort and discard its writes.
+            for addr in txn.participants.iter().cloned().collect::<Vec<_>>() {
+                self.send_abort(addr, &txn.id).await;
+            }
         }
```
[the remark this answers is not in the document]
JonathanAldrich: The AI description is incorrect, but there is an issue.  We should eliminate commit errors and only roll back when there is an execution error.

meerkat-lib/src/runtime/manager/mod.rs
JonathanAldrich: (retries will be implemented as part of asynchronous updates -- issue #24 -- though)

JonathanAldrich (changes requested):
Thanks for the PR!  This generally looks good (and the tests I tried worked for me) but there are a few things to look at.

Can you please add documentation on how to run the tests you added?  Or check out https://github.com/meerkat-lang/meerkat/pull/42/ (likely to merge this soon) and write a .json test that will be runnable with it.

Would it be possible to implement read locks on remote reads?  That seems like a relatively small addition that would remove the limitation and make arbitrary transactions work safely.  Otherwise we have to add it as a follow up issue; I don't mind that but maybe not to hard to just do it.

Please see comments for some other issues to address.

meerkat-lib/src/runtime/interpreter/executor.rs
```diff
@@ -32,10 +32,10 @@ pub async fn execute(
             match val {
                 Value::ActionClosure { stmts, env: closure_env, service_name: action_svc } => {
```
JonathanAldrich: One more thing--`ActionClosure` should be updated so that the `service_name` is instead a `ServiceId`.  Then `remote_action` should use the `addr` embedded in the `ActionClosure`'s `ServiceId` so the action will be executable even if it is not imported into the current scope.

JonathanAldrich: This is an easy change to make in this PR and should as a side effect address issue #40.

meerkat/src/main.rs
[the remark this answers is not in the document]
JonathanAldrich: yes, please open a GitHub issue for this

meerkat/src/main.rs
```diff
@@ -85,18 +86,15 @@ async fn run_server(
     port: u16,
     local: bool,
 ) -> Result<(), Box<dyn Error>> {
+    let mut net = NetworkActor::new(NodeType::Server).await?;
     let mut manager = Manager::new();
-    manager.local = local;
```
JonathanAldrich, line 89: I think this line needs to be present; without it, Dustin's mkn.py testing script breaks.

e.g. run 
`python3 scripts/mkn.py scripts/manifest1.json`

JonathanAldrich (changes requested):
Thank you for the revisions!  The code looks good now, however, there is a regression.  To see the problem, run Dustin's testing script

`python3 scripts/mkn.py scripts/manifest1.json`

I found a missing assignment in `main.rs` that fixes part of the problem but there is also an issue where `def`s are not updating properly anymore in some circumstances.

Related: can you add `.mkn` files to automate running the cross-node tests that you added?

meerkat/src/main.rs
dustinjuliano: They may also want to try `python3 scripts/test_mkn.py`, which will test both `mkn` (network orchestrator) and run a suite of examples to ensure it is working as intended; it also indirectly tests Meerkat!

JonathanAldrich (approved):
Everything looks good now!  Thanks!

PR #56, opened 2026-06-07
fix(ci): move codeql to no build mode

avishj:
CodeQL for Rust no longer requires a manual build step. It can directly use the `rust_analyzer` to analyze files without full compilation. This PR fixes CodeQL in Weekly Security by switching it to no build mode.

## Summary by CodeRabbit

* **Chores**
  * Updated security scanning workflow configuration to optimize code analysis operations.


JonathanAldrich (approved):
Thanks for the fix!

PR #57, opened 2026-06-08
feat: Update repl to use `rustyline` and support modern command-line features

andrewxzhou:
### Description
The current REPL lacks some modern command-line features (in particular, up-arrow/down-arrow to cycle through history and left-arrow/right-arrow to move the cursor). This PR updates the implementation of the repl in `repl.rs` to use the `rustyline` crate, which is a readline implementation that supports these features by default. 

### Changes:
 - Update `repl.rs` to use `rustyline` as the readline impl 
 - Get rid of previous stdio imports in `repl.rs` that were superseded by `rustyline`. 
 - Ctrl+C now exits the current prompt while staying in the REPL, while Ctrl+D remains the same in that it exits the REPL

### Additional notes
- I tried this out with some simple commands and it didn't seem to break anything, but I haven't done any robust testing yet. 

## Summary by CodeRabbit

* **New Features**
  * REPL now automatically saves and restores command history between sessions, allowing users to revisit previously executed commands for improved workflow continuity
  * Enhanced interactive input experience with improved keyboard interrupt handling for a more responsive REPL environment


meerkat/src/repl.rs
```diff
@@ -67,22 +67,25 @@ pub async fn run_repl(
...
+                reader.add_history_entry(l.as_str())?; l 
+            }
+            Err(ReadlineError::Interrupted) => {
+                buffer.clear();
+                continuation = false;
+                continue;
+            }
+            Err(ReadlineError::Eof) => break,
+            Err(e) => return Err(e.into()),
         };
```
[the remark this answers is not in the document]
andrewxzhou: I pushed a fix for the second and fourth issue so that there's now a warning when history is not saved and the repl supports multi-line history. In the fix for the fourth issue, I also made it so that history is saved on parse errors (eg if you typo a tiny bit it should still be saved). thank you big C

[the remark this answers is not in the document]
andrewxzhou: Also, I should note that I looked into the first and third issues briefly. The first issue (relative path for history) shouldn't be an issue for now since the REPL can only be run through a `cargo run`, which will always occur at the top level of the repo. This might be a concern later on though? 
And the third issue doesn't seem to be an actual issue - as far as I can tell all the rustyline examples I can find on the internet use this syntax anyways

andrewxzhou:
yippee checks pass!

meerkat/src/repl.rs
```diff
@@ -53,14 +54,12 @@ pub async fn run_repl(
     mut manager: Manager,
     remote_url_map: std::collections::HashMap<String, String>,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let stdin = io::stdin();
-    let is_tty = stdin.is_terminal();
+    let mut reader = DefaultEditor::new()?;
+    let _ = reader.load_history("meerkat_history.txt");
```
avishj: Should we be just storing this as a local file...or follow standard config/log file locations for each OS?

meerkat/src/repl.rs
```diff
@@ -143,9 +145,8 @@ pub async fn run_repl(
             }
         }
     }
-
-    if is_tty {
-        println!();
+    if let Err(e) = reader.save_history("meerkat_history.txt") {
```
avishj: Same concern as R58.

meerkat/src/repl.rs
```diff
@@ -53,14 +54,12 @@ pub async fn run_repl(
...
-    let is_tty = stdin.is_terminal();
+    let mut reader = DefaultEditor::new()?;
+    let _ = reader.load_history("meerkat_history.txt");
 
-    if is_tty {
-        println!("Meerkat REPL  (Ctrl-D to exit)");
-        println!("Enter service definitions, @test blocks, statements, or expressions.");
-        println!();
-    }
+    println!("Meerkat REPL  (Ctrl-D to exit)");
```
avishj: Can't say if this will ever be an issue, but maybe let's check and only print these on `tty`s so that it doesn't break in the case of `piped` inputs.

avishj (commented):
Couple of minor items, but overall should be good to go.

meerkat/src/repl.rs
```diff
@@ -116,11 +116,13 @@ pub async fn run_repl(
                 continuation = true;
             }
             ReplParseResult::Error(msg) => {
+                reader.add_history_entry(buffer.trim_end())?;
                 eprintln!("Parse error: {}", msg);
                 buffer.clear();
                 continuation = false;
             }
             ReplParseResult::Complete(stmts) => {
+                reader.add_history_entry(buffer.trim_end())?;
```
avishj: I think this could follow the same pattern as on R148, to nicely handle errors.

meerkat/src/repl.rs
```diff
@@ -116,11 +116,13 @@ pub async fn run_repl(
                 continuation = true;
             }
             ReplParseResult::Error(msg) => {
+                reader.add_history_entry(buffer.trim_end())?;
```
avishj: Same as R125.

avishj:
Great PR, I had missed a couple of items initially, added them later as comments.

meerkat/src/repl.rs
```diff
@@ -82,22 +81,23 @@ pub async fn run_repl(
...
+        };
 
-        let line = match lines.next() {
-            Some(Ok(l)) => l,
-            Some(Err(e)) => return Err(e.into()),
-            None => break,
+        let line = match readline {
+            Ok(l) => l,
+            Err(ReadlineError::Interrupted) => {
+                buffer.clear();
```
avishj, line 112: Oh, and also as a QOL thing, maybe print something like ^C or interrupted.

andrewxzhou:
added a commit to hopefully resolve these?

JonathanAldrich (changes requested):
Good idea!  A suggestion:
 * make the history file name start with `.` so it is hidden on unix platforms

JonathanAldrich:
Also add to .gitignore if the file is not guaranteed to go somewhere besides the project directory

Issue #43, opened 2026-06-03
Transactions - improving testing


avishj:
I think perhaps we could be testing a bit more than we currently do, and starting with transactions seems logical given that is complicated and can have various edge cases. This would be more focused on the internals rather than user facing mkt tests. With this in mind, here are some ideas on what we could start with - 

1. `VarLock` state machine unit tests - I think there can be a lot of edge cases here just like with any state machine. Some example invariants that we could test:
	- unlocked var accepts read lock.
	- unlocked var accepts write lock.
	- multiple read locks can coexist.
	- write lock blocks read lock from another transaction.
	- read lock blocks write lock from another transaction.
	- releasing the last read lock returns to unlocked.
	- releasing a write lock returns to unlocked.
	- releasing with the wrong transaction ID does not unlock.
2. `Manager::execute_action` unit tests to see commits/failures - Some ideas:
	- successful writes update both `value` and `latest_write_txn`.
	- failed transactions do not leave partial writes behind.
	- locks are released on both success and failure.
3. Classification of vars into read/write sets - I think it might also be useful to extract the classification logic from `execute_action_from_txn` and make it more modular. 
4. Regression testing for each bug found - Ideas from the 3 issues I just put out:	
    - failed transaction partially committing writes.
	- nested do writes missing latest_write_txn.
	- lookup reading through active write locks, IFF we decide that should not be allowed.

Issue #31, opened 2026-05-17
Acquire locks lazily during transactions


JonathanAldrich:
The current implementation of transactions acquires locks for every free variable in an action.  That won't work if an action can invoke another action returned by a function, because variables read and written in that other action are not free in the original action.  Furthermore, when we add `if` we only want to lock variables that are actually used.

Instead, locks should be acquired lazily as a transaction executes, when each variable is first read and written.  Acquiring the lock can be integrated with the read.  As an optimization, when a variable is read, the value should be cached in the transaction so it does not have to be read again, thus avoiding extra network traffic for the read if it is remote.

PR #61, opened 2026-06-10
feat: add testing for txn and execute action

avishj:
Completes #43 part 1, 2. Part 3, 4 are no relevant since lazy locking makes having static read/write sets unnecessary and some of the bugs were intended behavior.

Signed-off-by: Avish Jha <avish.j@pm.me>

## Summary by CodeRabbit

* **Tests**
  * Added test helpers and expanded coverage for transaction commit and failure scenarios, ensuring successful commits update state while failed transactions do not override prior commits and locks are released.
  * Verified lock state management: read/write lock semantics, multiple-reader coexistence, blocking between readers and writers, proper unlock/upgrade behavior, and idempotent/invalid-upgrade handling.


avishj:
@coderabbitai full review

avishj:
@coderabbitai full review

ShreejithBabuG (approved):
Reviewed and approving. 

Coverage for parts 1 and 2 is thorough. The VarLock tests exercise the full state machine: read and write acquisition, multiple-reader coexistence, blocking in both directions, release semantics including wrong-transaction-id no-ops, and all four upgrade cases. The execute_action tests confirm that a successful commit updates both value and latest_write_txn, a failed transaction preserves the previously committed state, and locks are released on both the success and failure paths.

On part 3: since locks are acquired lazily during execution (#31), there is no static read/write set to extract, so dropping it is correct.

On part 4, I verified the three items against the transaction code:

Failed transactions leaving partial writes: covered by the rollback test.

Nested do missing latest_write_txn: confirmed it is recorded. A nested do reuses the same transaction, so its buffered write is stamped at commit like any other.

Lookup reading through an active write lock: confirmed that a non-transactional read returns the committed value. This is intended, since the uncommitted write is held in the transaction buffer and applied only at commit, so a plain read cannot observe an in-flight write.

Both of these follow from the buffer-then-commit design, so I am comfortable treating them as intended behaviour.

One optional, non-blocking follow-up: asserting latest_write_txn in the existing nested-do test would lock that behaviour in explicitly. Approving as-is.

PR #65, opened 2026-06-11
refactor(meerkat-lib): resolve compiler warnings and clean up lints

dustinjuliano:
Meerkat has warnings in its current workspace build. This spams logs, tests, and builds and also is generally not recommended to ignore; I try to treat warnings as errors for strong maintainability.

There were also some warning suppressions, which I caution against, as it disables the purpose of warnings and reduces safety. These have all been eliminated and the warnings they would have generated, corrected, workspace wide.

The tokenizer and parser generator were left untouched for this audit, which need their suppressions due to the way they often generate boilerplate code.

I ran local hooks and CI. All tests passing. Ready for review and merge.

### DIFF
--- a/meerkat-lib/src/runtime/var_actor/handler.rs
+++ b/meerkat-lib/src/runtime/var_actor/handler.rs
@@ -164,22 +164,6 @@ impl kameo::prelude::Message<Msg> for VarActor {
 
                 Msg::Unit
             }
-
-            Msg::TestRequestPred { from_mgr_addr, test_id } => {
-                info!("Only for asserts: Pred Request from {:?}", from_mgr_addr);
-
-                // will immediately send back latest pred id
-                let _ = from_mgr_addr.tell(
-                    Msg::TestRequestPredGranted { 
-                        from_name: self.name.clone(),
-                        test_id,
-                        pred_id: self.latest_write_txn.clone().map(|t| t.id) 
-                }).await;
-
-                Msg::Unit
-            }
-
-            #[allow(unreachable_patterns)]
             _ => panic!("VarActor should not receive message {:?}", msg),
         }
     }
--- a/meerkat-lib/src/error.rs
+++ b/meerkat-lib/src/error.rs
@@ -1,5 +1,3 @@
-use std::fmt::Display;
-use std::io;
 use std::result;
 
 pub type Result<T> = result::Result<T, Error>;
--- a/meerkat-lib/src/runtime/ast/mod.rs
+++ b/meerkat-lib/src/runtime/ast/mod.rs
@@ -222,10 +222,7 @@ impl Display for Value {
             Value::Bool { val } => write!(f, "{}", val),
             Value::String { val } => write!(f, "\"{}\"", val),
             Value::Closure {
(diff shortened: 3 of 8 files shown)


meerkat-lib/src/runtime/var_actor/handler.rs
```diff
@@ -164,22 +164,6 @@ impl kameo::prelude::Message<Msg> for VarActor {
 
                 Msg::Unit
             }
-
-            Msg::TestRequestPred { from_mgr_addr, test_id } => {
```
andrewxzhou, line 168: Just wondering why this was removed? I tried to add it back (without the unreachable pattern macro) and no errors were produced

andrewxzhou (commented):
looked at everything and everything appears to be good except for the deletion I commented on

meerkat-lib/src/runtime/var_actor/handler.rs
dustinjuliano: Thank you for the review. I will explain the removal of that block:

In the original file from line 62 to 74 that block exists identically as:

```
Msg::TestRequestPred { from_mgr_addr, test_id } => {
    info!("Only for asserts: Pred Request from {:?}", from_mgr_addr);

    // will immediately send back latest pred id
    let _ = from_mgr_addr.tell(
        Msg::TestRequestPredGranted { 
            from_name: self.name.clone(),
            test_id,
            pred_id: self.latest_write_txn.clone().map(|t| t.id) 
    }).await;

    Msg::Unit
}
```

Then from line 168 to 180:

```
Msg::TestRequestPred { from_mgr_addr, test_id } => {
    info!("Only for asserts: Pred Request from {:?}", from_mgr_addr);

    // will immediately send back latest pred id
    let _ = from_mgr_addr.tell(
        Msg::TestRequestPredGranted { 
            from_name: self.name.clone(),
            test_id,
            pred_id: self.latest_write_txn.clone().map(|t| t.id) 
    }).await;

    Msg::Unit
}
```

These match arms are identical and one of them was removed so that the `#[allow(unreachable_patterns)]` warning surpression could be removed. Adding it back is not an error, but it should generate a warning, which is the goal of this PR to eliminate.

Also, I found out that this whole folder is dead code and is not built into the project, and we may want to do another PR to clean it up later in the future. Appears to have become dead code around March 9th, 2026 by commit 2277118.

This is the reason you did not see any warnings produced when you reintroduced the duplicate match arm, as this file is currently not part of the build.

In all cases, this deletion is safe and removed block of code is actually present starting on line 64, so the match arm logic is still there.

andrewxzhou: Oh that's really funny. In that case everything else LGTM

dustinjuliano:
Thank you for the review. All review comments have been addressed.

Edit: This PR also fixes some small `rustfmt` regressions introduced by PR #60, which was because that was authored before the transition to `rustfmt`. This PR will bring the project to baseline Rust format compliance in addition to removing the warnings and supressions.

JonathanAldrich (approved):

Issue #34, opened 2026-05-26
type checking for self-contained programs


JonathanAldrich:
Get type checking working for self-contained programs (those with no imports).  There is type checking code that was working for a previous version that had (mostly) the same AST, but is not currently running; that can be a starting point.

dustinjuliano:
On 2026-05-28 one-on-one meeting with Professor Aldrich we decided to go for bi-directional typing system, away from the HM code that is present (but inactive/disconnected). Future consideration is cycle detection/prevention for dependency graph to be added in a future round of updates after this initial static type checking is operational.

dustinjuliano:
2026-06-10 Status Update

- PR #37 was initially opened as a draft to work on this issue. I proposed in a group meeting on 2026-06-09 to split that PR into multiple dependent ones to adhere to atomic commit strategy, streamlined reviews, and to make it easier to keep synchronized as multiple developers are working on sometimes overlapping parts of the project.
- By this time I had already implemented several prerequisite features leading up to the bi-directional type checking algorithm. Due to my own fault, some of these commits were overlapping and were not atomic. I took time to perform significant interactive staging and interactive rebasing to cherry-pick and clean things up on my end. I then introduced explicit feature branches in a hierarchical order and then closed the original PR #37 that was associated with the obsolete `static-types` branch on my end.
- The new hierarchy of feature branches on my fork look like this, and they are all related to solving Issue 34:
	- `refactor-del-hm-type-system`
		- `feat-node`
			- `feat-check-mode-cli`
				- `feat-ast-printer`
					- `feat-string-interning`
						- `feat-static-name-resolution`
- The branch `refactor-del-hm-type-system` is the head branch for PR #60, and its base branch is `main`.
- Currently, I believe it is easiest on Professor @JonathanAldrich  and other team members if I incrementally introduce PR requests as the parent branches are accepted and merged. This also allows me to remain flexible if I need to refactor or reorganize dependencies on my fork and avoids spurious PR requests and comments from Code Rabbit and CI/CD as the drafting process unfolds.
- All of the above feature branches, except for `feat-static-name-resolution` (which I am working on now), are ready to be reviewed and merged into `main` once their ancestor/parent branches are merged.
- This workflow, which I researched extensively, allows me to create atomic commits and build and branch in a dependent way that makes each PR an isolated/separable concern that is minimal and easy to review, with very clean commit history. However, it does present some challenges during the initial transition:
[...]

PR #70, opened 2026-06-14
feat: add `--check` CLI for static check only mode

dustinjuliano:
Adds a `--check` CLI that is intended to run static checks and then terminate without running the program. It is currently a stub. This option makes the edit-test cycle much faster. Name resolution and type checking features depend on this PR for work on Issue #34.

## Summary by CodeRabbit

* **New Features**
  * Added a `--check` command-line flag to parse and validate input files without executing server mode, client operations, service loading, or starting the interactive REPL.


JonathanAldrich (approved):
Good idea

Issue #62, opened 2026-06-11
bug: non-deterministic hang in circuit relay test


dustinjuliano:
Running `cargo test` would sometimes hang. I isolated it to the circuit relay test. During investigation I found a possible bug in the runtime in addition to the test itself.

1. The test is using simple timers and may want to use events for tests running in parallel or pressure on `tokio` that manifests the non-determinism, but also affected by (2)
2. The runtime bug appears to occur because the event loop silently drops the asynchronous dialing and listener failures instead of handling them

dustinjuliano:
A fix for this has been proposed in PR #69.

dustinjuliano:
I keep running into issues with the circuit relay test fixes. To respond better, I expanded my analysis to `rust-libp2p` itself and found some important context. I did an audit to find some issues, past and present, related to the circuit relay feature of `rust-libp2p`. To save you from reading the entire comment threads, I had AI summarize them:

- `rust-libp2p` [Issue 6165](https://github.com/libp2p/rust-libp2p/issues/6165)
	- Core Proposal: The issue describes a bug where closing a listener tied to a circuit relay reservation does not cleanly remove the associated external relayed multiaddress from the node's local records, meaning the `Swarm` fails to signal that the address has expired
	- Mechanism: In `protocols/relay/src/priv_client.rs`, the relay client logic only emits the `ToSwarm::ExternalAddrExpired` event when the underlying connection itself breaks. It lacks a direct cleanup step when the listener alone is shut down manually or by the transport
	- Proposed Requirements: The client needs a safe mechanism to signal that the listener closed and expire the external address, proving the node is no longer reachable through that specific relay route
	- Architecture Implementation: Fixing the bug directly inside `on_listener_closed` is blocked by an edge case. A framework test (`new_reservation_to_same_relay_replaces_old`) dictates that if a new reservation replaces an old one on the exact same relay, the old listener closes but the address must persist through the replacement handover flow to keep the node reachable
	- Current Status: The issue is Closed as completed. It was resolved via Pull Request `#6285` (`fix(relay): remove external addr if there is no replacement reservation`), which introduced a conditional cleanup logic that only expires the address if a handover reservation is not active

- `rust-libp2p` [Issue 6238](https://github.com/libp2p/rust-libp2p/issues/6238)
[...]

PR #71, opened 2026-06-14
ci: enforce warnings as errors across all cargo builds

dustinjuliano:
Add the `RUSTFLAGS: "-D warnings"` environment variable globally to the Rust CI job in `ci.yml`. This ensures that all `cargo check`, `cargo test`, `cargo run`, and `cargo build` steps treat compiler warnings as errors, covering 100% of targets, modules, and packages in the workspace (including the WASM target).

This prevents warnings from going unnoticed in pull requests. Local verification confirms that the workspace builds cleanly without warnings under this configuration.

(This will not affect local edit-test cycles; it is only part of the CI check. It is intended to safeguard `main` branch. This should be highly compatible with linters, which would generally do the same.)

## Summary by CodeRabbit

* **Chores**
  * Enhanced internal build quality assurance processes to enforce stricter compilation standards.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     permissions: {}
+    env:
+      RUSTFLAGS: "-D warnings"
     steps:
       # Monitors outbound network calls and can enforce egress policy. Audit mode
       # records unexpected traffic without blocking while CI hardening is tuned.


dustinjuliano:
The CI has failed due to a test failure in `test_multiple_messages`. I am investigating. There should be no reason that this PR would have caused that to fail. This is likely non-determinism manifesting again related to Issue #62 (currently closed, but may need to be reopened).

dustinjuliano:
I have confirmed that the CI failure is 100% unrelated to this PR. The cause is indeed a non-deterministic bug in the way `libp2p` is being used currently. I will open an issue with details.

This PR can be reviewed and merged independently of this new finding.

JonathanAldrich (approved):
Good idea

PR #77, opened 2026-06-15
fixing an off-by-one error found by coderabbitai

JonathanAldrich:
See discussion in https://github.com/meerkat-lang/meerkat/pull/74 (coderabbitai's review with an actionable comment outside diff range)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected numeric bounds validation during integer parsing to properly reject out-of-range number literals.


dustinjuliano (approved):
This looks like a strong fix. I found in analysis that Meerkat treats negative literals as a sequence of unary negation operator and a numeric literal that is always non-negative. This change makes it safe to always negate any value `Token::Number` for it will be in the valid range of `i32`.

Issue #45, opened 2026-06-03
Service initialization should be done in a transaction


JonathanAldrich:
This will ensure that the fields and defs of the service are initialized consistently.  One transaction should be used to initialize all fields and defs.

PR #87, opened 2026-06-20
feat: Initialize services in `create_service` with a single transaction

andrewxzhou:
### Description
Previously, service initialization in `create_service` was done entirely without the use of transactions. Presumably this could lead to race conditions and other concurrency bugs if multiple nodes tried to create a service at the same time, as stated by issue #45. This PR implements that.

### Changes
- Add a transaction to `create_service`
- `eval` within `create_service` now uses that transaction
- Add error handling for transaction failures along the pattern of `execute_action_with_txn`
- Add code to acquire/release locks
- Make it so `latest_write_txn` records the transaction of `create_service`
- Add some test cases

### Additional Info/Notes for review
To be quite honest I think I still don't fully understand the transaction model and codebase. That said, there are a few things that stand out to me as suspicious.
- On lines 242 and 246, I'm not sure if it's still correct to just push the variable value directly onto the environment and the service.vars, or if this also needs to go behind a transaction somehow for concurrency purposes. There's also a symmetric case for this for definitions.
- in the test case `test_txn_read_lock_released_after_failure`, there's an assertion that the `latest_write_txn` `is_none()`. I suspect that, since we now have `create_service` use a txn, this should be updated to check that the latest txn is the same as the one used by `create_service`. This currently produces a test case failure, so I'd greatly appreciate knowing if the test case is wrong or if I'm wrong.
- Speaking of test cases, I'm honestly not sure how to test that the transaction does what is advertised (i.e. it makes it so service initialization is fully atomic). To be completely honest I let Claude write the test cases, and I believe the last test case it wrote, `test_create_service_read_conflicts_under_one_txn`, tries to test this. However, I don't fully understand the concurrency stuff, so I'm not sure if this test case is robust and actually tests this behavior. (The other two test cases in this PR were also written by Claude, but at least for those I've gone through them fully and convinced myself that they are robust and test what they claim to test. This is very much not the case for the last test case). 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service initialization so all variable/definition initializers run under a single shared transaction for atomic behavior.
  * Initialization errors now trigger a clean abort/rollback flow (best-effort abort, remove partially created service), while successful initialization commits and records the same transaction consistently.
  * Initialization locks are now released reliably after either success or failure.

* **Tests**
  * Updated and added coverage for transactional initialization behavior, rollback correctness, service removal on partial failures, and lock-conflict handling.

### DIFF
--- a/meerkat-lib/src/runtime/manager/mod.rs
+++ b/meerkat-lib/src/runtime/manager/mod.rs
@@ -2336,7 +2336,7 @@ mod tests {
                 is_pub: true,
             },
         ];
-        tc.manager.create_service(tc.f, decls).await.unwrap();
+        tc.manager.create_service(tc.foo, decls).await.unwrap();
 
         let foo = tc.manager.services.get(&tc.foo).unwrap();
         let tid = foo.vars.get(&tc.x).unwrap().latest_write_txn.clone();
--- a/meerkat-lib/src/runtime/manager/mod.rs
+++ b/meerkat-lib/src/runtime/manager/mod.rs
@@ -231,35 +231,50 @@ impl Manager {
         let mut env: Vec<(Symbol, Value)> = vec![];
         let svc_name = name;
 
+        let mut txn = Transaction::new(TxnId::new(self.node_id));
+        let mut init_error = None;
+
         for decl in decls {
             match decl {
                 Decl::VarDecl { name, val } => {
-                    let value = eval(
+                    let value = match eval(
                         &val,
                         &env,
                         &mut EvalContext {
                             manager: self,
                             service_name: svc_name,
-                            txn: None,
+                            txn: Some(&mut txn),
                         },
                     )
-                    .await?;
+                    .await {
+                        Ok(v) => v,
+                        Err(e) => {
+                            init_error = Some(e);
+                            break;
+                        }
+                    };
                     env.push((name, value.clone()));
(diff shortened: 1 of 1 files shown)


ShreejithBabuG:
Thanks for the honest write-up. I read through the diff, and it is in good shape, your instincts were sound.

The structure is right, and the part that actually makes it atomic is the failure path removing the whole service with 'services.remove' . Since the service did not exist before, that is a clean rollback.

On the direct 'env.push' and 'service.vars.insert', both are fine. 'env' is the local scratchpad, and because failure removes the service you get all-or-nothing without buffering. The only nuance is that those writes are not behind write locks, so they are not isolated mid-init, but the service is not reachable until 'create_service' returns, so it is low risk.

Your fix to the 'latest_write_txn' assertion is correct. The open question, whether init should record 'latest_write_txn' at all, is a design call for Professor @JonathanAldrich , since a completed init txn now shows up as the last writer.

The atomicity test is more robust than you thought, it properly exercises the lock conflict and the rollback. Two small additions would tighten it: a comment noting it relies on s2's txn being younger (so it dies rather than waits), and a success-path test where a committed init becomes visible to another transaction only after commit.

Overall this looks good. The only thing to settle before merge is the 'latest_write_txn' semantics, which is his call.

JonathanAldrich (approved):
Looks good.  I concur with setting the last transaction to the one that initializes the service.  Thanks for the PR!

Issue #80, opened 2026-06-16
RFC on String Interning


dustinjuliano:
I am writing this to document my work, ask questions, and synchronize with team members who may be affected by this upcoming feature.

For background context, the chain of prerequisites flows like this, leading up to static type checking:

- String Interning
- Static Name Resolution
- Static Type Checking

It is possible to implement name resolution and type checking entirely with strings. However, I would recommend against it.
## String/Symbol IDs

Adding string interning is straightforward. However, making this change would involve updating every use of a Rust `String` that refers to names with `Symbol`s, which are defined like this:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
pub struct Symbol(pub u32);
```

There is a rationale for why the `Symbol` is represented by `u32` and not something else. During static name resolution there is a need to construct environments. There are significant benefits if these can be encoded in a single `u64`. That gives another `u32` worth of information to make data structure designs that can have very impactful reductions on time and space. This is an implementation detail that allows for an efficient symbol table construction.

## Binding IDs

There is also a tentative plan to introduce another data type called `BindingId`:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
pub struct BindingId(pub u64);
```

The choice of `u64` for its representation is important because we would be relying on the enormous value range of that type to avoid having to maintain free lists for `BindingId`s that are no longer used when code that refers to them is dropped.

Unlike `Symbol`, `BindingId`s are unique to the global namespace of a local Meerkat node (for clarity: they are not network unique, just node unique). To see the difference between them, consider this pseudocode fragment:

```rust
let x = 0;   // "x" => Symbol(1), BindingId(1)
let x = 0;   // "x" => Symbol(1), BindingId(2)
```

The second `x` has a distinct `BindingId`, but the exact same `Symbol`. This has ramifications when the use of a name needs to be distinguished from just the name itself. Having unique identifiers for the binding site of a name has numerous benefits in the static and dynamic semantics. There should be no issue at the network edge, as Meerkat currently transmits names as strings, which can be interned for internal use within a node; a distributed or network-wide symbol table is not required.

Another benefit that `BindingId` would confer is when it comes to closures and dynamic name resolution. It could potentially be used to avoid a walk up a parent-pointer structure to resolve names.

[...]

PR #90, opened 2026-06-22
Asynchronous change-notification propagation (#24)

nivethadhanakoti:
Replaces the polling Manager::propagate with push-based change notifications, per the Historiographer "Simple protocol", closing #24. When a def initializes, it registers as a listener on the members it depends on; when a member changes, the manager notifies each listener (local in-process, remote over the wire); the listener recomputes from cached dependency values without a round-trip.

**Changes made**
- Expr::cross_service_deps extracts the (service, member) references a def depends on (the counterpart to free_var, which drops MemberAccess).
- Service gains listeners, dep_cache, and dep_remote; create_service registers same-service listeners (dep_graph) and cross-service ones (local owners in-process, remote owners via subscribe).
- propagate is rewritten as a (service, member) worklist over the listener graph, replacing the topo-scan recompute.
- New RequestUpdates / Update messages, with handle_request_updates (owner registers a listener and replies with the current value) and handle_update (listener caches the value and recomputes).
- A transient reactive cache on Manager: MemberAccess resolves from cache during a recompute, so a change does not trigger a fresh lookup.
- A --watch CLI flag that prints change notifications as they arrive, for every dependency, without a REPL.

**Design notes**
Symbols are interner-local, so the wire carries names as strings (interned on receipt) and values via the existing NetValue codec. A listener identifies itself by its globally-unique ServiceNetId, which the owner echoes back so the listener can match it to a local service.

**Testing**
- Unit tests (75 lib, 10 integration green): cross-service dep extraction, eager cross-service cascade, handle_update recompute-from-cache (asserts 12 from the cached push rather than 4 from a local lookup, proving no round-trip), and listener registration.
- Manual three-terminal demo with s1.mkt / test_client.mkt: an s1 server, a --watch client on s2, and a trigger client invoking s1.inc_x. The watcher prints the s1.y change and the recomputed z over the wire.
- Honest gap: the wire path is covered by the manual demo plus manager-level unit tests, not yet by an automated two-node integration test in CI. Happy to add one if preferred.

**Follow-ups**

- A def that references a remote action member (here inc_delegate references s1.inc_x) also subscribes to that action member, so the watcher prints one extra initial notification for it. An action closure never changes, so this is noise rather than a correctness issue; filtering action members at registration would remove it.
- A one-shot client that subscribes and then exits leaves a stale listener address on the owner. Dropping listeners on send failure or disconnect is a reasonable follow-up.

## Release Notes

* **New Features**
  * Added `--watch` mode to monitor distributed reactive changes and continuously print updated listener values.
  * Introduced subscription-style cross-service listener updates to enable eager recomputation and remote propagation when members change.

* **Performance**
  * Improved member-access evaluation during reactive recomputation by using a short-lived cache to reduce redundant resolutions.

* **Behavior/UX**
  * Updated networking to support listener subscription requests and pushed update delivery, while keeping watched listener values current.

### DIFF
--- a/meerkat-lib/src/net/codec.rs
+++ b/meerkat-lib/src/net/codec.rs
@@ -8,7 +8,7 @@ use crate::net::ast::{
     NetActionStmt, NetBinOp, NetExpr, NetField, NetParam, NetTableType, NetType, NetUnOp, NetValue,
 };
 use crate::runtime::ast::{ActionStmt, BinOp, Expr, Field, TableType, UnOp, Value};
-use crate::runtime::interner::Interner;
+use crate::runtime::interner::{Interner, Symbol};
 use crate::runtime::limits::{MAX_IDENTIFIER_LENGTH, MAX_STRING_LITERAL_LENGTH, MAX_TYPE_DEPTH};
 use crate::runtime::tt::{Param, TupleType, Type};
 
@@ -30,6 +30,51 @@ fn validate_string_literal(s: &str) -> Result<()> {
         )));
     }
     Ok(())
+}
+
+/// #24: validate and intern the identifier fields of a `RequestUpdates`
+/// message arriving over the wire. Per the zero-trust boundary, all interning
+/// of network-sourced names happens here in codec (never in the manager or
+/// main), after `validate_identifier`. `listener_service` and `reply_to` are
+/// network addresses, not identifiers, so they are not interned here.
+///
+/// Returns `(service, member, listener_def)` as interned `Symbol`s.
+pub fn decode_request_updates(
+    service: &str,
+    member: &str,
+    listener_def: &str,
+    interner: &mut Interner,
+) -> Result<(Symbol, Symbol, Symbol)> {
+    validate_identifier(service)?;
+    validate_identifier(member)?;
+    validate_identifier(listener_def)?;
+    Ok((
+        interner.insert(service),
+        interner.insert(member),
+        interner.insert(listener_def),
+    ))
+}
+
+/// #24: validate and intern the identifier fields of an `Update` message
+/// arriving over the wire (the `value` is decoded separately via
+/// `decode_value`). As with `decode_request_updates`, all interning of
+/// network-sourced names happens here after validation.
+///
+/// Returns `(listener_def, source_service, member)` as interned `Symbol`s.
+pub fn decode_update(
+    listener_def: &str,
+    source_service: &str,
+    member: &str,
+    interner: &mut Interner,
+) -> Result<(Symbol, Symbol, Symbol)> {
+    validate_identifier(listener_def)?;
+    validate_identifier(source_service)?;
+    validate_identifier(member)?;
+    Ok((
+        interner.insert(listener_def),
+        interner.insert(source_service),
+        interner.insert(member),
+    ))
 }
 
 /// Encode a runtime `Type` using recursion depth accumulator
--- a/meerkat-lib/src/net/codec.rs
+++ b/meerkat-lib/src/net/codec.rs
@@ -19,6 +19,23 @@ fn validate_identifier(s: &str) -> Result<()> {
             MAX_IDENTIFIER_LENGTH
         )));
     }
+    // Match the lexer's identifier rule ([A-Za-z_][A-Za-z0-9_]*) so a name
+    // arriving over the wire is rejected unless the parser would also accept
+    // it locally. Without this, length-valid but non-identifier strings
+    // (empty, whitespace, punctuation) could still be interned from the wire.
+    let mut chars = s.chars();
+    let valid = match chars.next() {
+        Some(c) if c.is_ascii_alphabetic() || c == '_' => {
+            chars.all(|c| c.is_ascii_alphanumeric() || c == '_')
+        }
+        _ => false,
+    };
+    if !valid {
+        return Err(Error::LimitExceeded(format!(
+            "invalid identifier (must match [A-Za-z_][A-Za-z0-9_]*): {:?}",
+            s
+        )));
+    }
     Ok(())
 }
 
@@ -1319,6 +1336,31 @@ mod tests {
         let res = decode_value(net_val, &mut interner);
         assert!(res.is_err());
         assert!(matches!(res.unwrap_err(), Error::LimitExceeded(_)));
+    }
+
+    /// #24: verify the codec rejects wire identifiers that the lexer would not
+    /// accept locally (whitespace, punctuation, leading digit, empty), so a
+    /// remote peer cannot grow the interner with non-identifier strings.
+    #[test]
+    fn test_codec_decode_rejects_non_identifier() {
+        let mut interner = Interner::new();
+        for bad in ["bad name", "bad!name", "1leading", "", "has.dot", "sp ace"] {
+            let res = decode_request_updates(bad, "member", "ldef", &mut interner);
+            assert!(
+                res.is_err(),
+                "decode_request_updates should reject service {:?}",
+                bad
+            );
+            let res = decode_update("ldef", bad, "member", &mut interner);
+            assert!(
+                res.is_err(),
+                "decode_update should reject source_service {:?}",
+                bad
+            );
+        }
+        // A well-formed identifier is accepted.
+        assert!(decode_request_updates("svc", "y", "z", &mut interner).is_ok());
+        assert!(decode_update("z", "svc", "y", &mut interner).is_ok());
     }
 
     /// Verify that encoding a value with an oversized string
--- a/meerkat-lib/src/net/types.rs
+++ b/meerkat-lib/src/net/types.rs
@@ -139,6 +139,30 @@ pub enum MeerkatMessage {
     /// request is alive and still queued, so it keeps waiting
     /// instead of timing out
     WaitParked { request_id: u64 },
+
+    /// #24: subscribe `listener_service.listener_def` (living on the sender's
+    /// node) as a listener on `service.member` (on this node). The owner
+    /// registers it and replies with an initial `Update` carrying the current
+    /// value. Names are strings; symbols are interner-local.
+    RequestUpdates {
+        request_id: u64,
+        service: String,
+        member: String,
+        listener_service: String,
+        listener_def: String,
+        reply_to: String,
+    },
+
+    /// #24: tell a remote listener that `source_service.member` changed (also
+    /// used for the initial value on subscribe). The receiver caches `value`
+    /// and recomputes `listener_def`.
+    Update {
+        listener_service: String,
+        listener_def: String,
+        source_service: String,
+        member: String,
+        value: NetValue,
+    },
 }
 
 /// Errors that can occur when sending messages
--- a/meerkat-lib/src/net/types.rs
+++ b/meerkat-lib/src/net/types.rs
@@ -143,7 +143,10 @@ pub enum MeerkatMessage {
     /// #24: subscribe `listener_service.listener_def` (living on the sender's
     /// node) as a listener on `service.member` (on this node). The owner
     /// registers it and replies with an initial `Update` carrying the current
-    /// value. Names are strings; symbols are interner-local.
+    /// value. Services, members, and defs are carried as strings rather than
+    /// symbols: symbols are interner ids local to one node, but these messages
+    /// cross nodes, so each side must resolve the names through its own interner
+    /// to know which service/member the request refers to.
     RequestUpdates {
         request_id: u64,
         service: String,
--- a/meerkat-lib/src/runtime/interpreter/evaluator.rs
+++ b/meerkat-lib/src/runtime/interpreter/evaluator.rs
@@ -255,6 +255,18 @@ pub async fn eval(
             service_name,
             member_name,
         } => {
+            // #24: during a reactive recompute the cross-service deps are already
+            // cached, so resolve from the cache and skip the lookup (which for a
+            // remote service would be a network round-trip).
+            if let Some(v) = ctx
+                .manager
+                .reactive_cache
+                .as_ref()
+                .and_then(|c| c.get(&(service_name, member_name)))
+                .cloned()
+            {
+                return Ok(v);
+            }
             // The `Manager` determines whether the service is local or
             // remote
             ctx.manager
--- a/meerkat-lib/src/runtime/interpreter/evaluator.rs
+++ b/meerkat-lib/src/runtime/interpreter/evaluator.rs
@@ -255,9 +255,10 @@ pub async fn eval(
             service_name,
             member_name,
         } => {
-            // #24: during a reactive recompute the cross-service deps are already
-            // cached, so resolve from the cache and skip the lookup (which for a
-            // remote service would be a network round-trip).
+            // #24: during a reactive update we check the cache first. If this
+            // (service, member) was already fetched for the def being recomputed,
+            // use the cached value instead of doing a lookup (which for a remote
+            // service would be a network round-trip).
             if let Some(v) = ctx
                 .manager
                 .reactive_cache
--- a/meerkat-lib/src/runtime/manager/mod.rs
+++ b/meerkat-lib/src/runtime/manager/mod.rs
@@ -97,6 +97,11 @@ pub struct Manager {
     /// #24: reply address for each remote listener, keyed by the listener's
     /// ServiceNetId, so the owner can route Updates back to it.
     pub listener_addrs: HashMap<ServiceNetId, String>,
+    /// #24: RequestUpdates/Update messages seen by the sync reply-router
+    /// (dispatch_network_events) while awaiting a reply. They carry no
+    /// request_id, so they are buffered here and drained in async context so a
+    /// reactive update arriving mid-wait is applied, not dropped.
+    pub pending_reactive: Vec<MeerkatMessage>,
 }
 
 impl Manager {
@@ -114,6 +119,7 @@ impl Manager {
             interner,
             reactive_cache: None,
             listener_addrs: HashMap::new(),
+            pending_reactive: Vec::new(),
         }
     }
 
@@ -839,6 +845,15 @@ impl Manager {
             let event = n.try_recv_event();
             match event {
                 Some(NetworkEvent::MessageReceived { msg, .. }) => {
+                    // #24: these carry no request_id and are not replies; buffer
+                    // them for async draining instead of dropping them here.
+                    if matches!(
+                        msg,
+                        MeerkatMessage::RequestUpdates { .. } | MeerkatMessage::Update { .. }
+                    ) {
+                        self.pending_reactive.push(msg);
+                        continue;
+                    }
                     let rid = match &msg {
                         MeerkatMessage::LookupResponse { request_id, .. } => Some(*request_id),
                         MeerkatMessage::LookupError { request_id, .. } => Some(*request_id),
@@ -873,6 +888,53 @@ impl Manager {
     }
 
     /// shared by remote_lookup and remote_action.
+    /// #24: handle any reactive messages the sync router buffered. Called in
+    /// async context (after each dispatch_network_events) so RequestUpdates and
+    /// Update arriving mid-wait are applied rather than lost.
+    async fn drain_pending_reactive(&mut self) {
+        while !self.pending_reactive.is_empty() {
+            let batch = std::mem::take(&mut self.pending_reactive);
+            for msg in batch {
+                match msg {
+                    MeerkatMessage::RequestUpdates {
+                        service,
+                        member,
+                        listener_service,
+                        listener_def,
+                        reply_to,
+                        ..
+                    } => {
+                        self.handle_request_updates(
+                            service,
+                            member,
+                            listener_service,
+                            listener_def,
+                            reply_to,
+                        )
+                        .await;
+                    }
+                    MeerkatMessage::Update {
+                        listener_service,
+                        listener_def,
+                        source_service,
+                        member,
+                        value,
+                    } => {
+                        self.handle_update(
+                            listener_service,
+                            listener_def,
+                            source_service,
+                            member,
+                            value,
+                        )
+                        .await;
+                    }
+                    _ => {}
+                }
+            }
+        }
+    }
+
     async fn send_and_await_reply(
         &mut self,
         addr: Address,
@@ -900,6 +962,7 @@ impl Manager {
 
         loop {
             self.dispatch_network_events();
+            self.drain_pending_reactive().await;
             tokio::select! {
                 biased;
                 result = &mut rx => {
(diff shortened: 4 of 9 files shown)


meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -448,76 +519,262 @@ impl Manager {
...
+            None => false,
+        }
+    }
+
+    /// #24: fire-and-forget send (no reply awaited).
+    async fn send_oneway(&mut self, addr: Address, msg: MeerkatMessage) {
+        if let Some(net) = self.network.as_mut() {
+            net.handle_command(NetworkCommand::SendMessage { addr, msg })
+                .await;
+        }
```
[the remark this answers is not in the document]
JonathanAldrich: This would be good to fix, either in this PR or in a future issue.  Basically remove a listener from the list if there is a failure on a send (or perhaps several failures).  We don't want to block on the send but an async hook that does this in the failure case would be good.

In the longer term, shutting down a node should also remove all the subscriptions to remote vars/defs that it has.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -448,76 +519,262 @@ impl Manager {
...
+        let member_sym = self.interner.insert(&member);
+        let listener_def_sym = self.interner.insert(&listener_def);
+        let listener_id = ServiceNetId(listener_service.clone());
+
+        if let Some(svc) = self.services.get_mut(&service_sym) {
+            svc.listeners
+                .entry(member_sym)
+                .or_default()
+                .insert((listener_id.clone(), listener_def_sym));
+        } else {
```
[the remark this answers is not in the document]
JonathanAldrich: good suggestion

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -540,7 +797,9 @@ impl Manager {
                         | MeerkatMessage::LookupRequest { .. }
                         | MeerkatMessage::ActionRequest { .. }
                         | MeerkatMessage::Commit { .. }
-                        | MeerkatMessage::Abort { .. } => None,
+                        | MeerkatMessage::Abort { .. }
+                        | MeerkatMessage::RequestUpdates { .. }
+                        | MeerkatMessage::Update { .. } => None,
                     };
```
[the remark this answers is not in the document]
JonathanAldrich: @nivethadhanakoti @ShreejithBabuG can you investigate whether this is a real issue and fix if so?

meerkat/src/main.rs
```diff
@@ -529,6 +572,124 @@ async fn run_server(
...
+    if let NetworkReply::ListenSuccess { addr } = reply {
+        let node_ip = manager.get_node_ip();
+        let peer_id = net.local_peer_id();
+        let addr_str = addr
+            .0
+            .replace("0.0.0.0", &node_ip)
+            .replace("127.0.0.1", &node_ip);
+        manager.set_local_address(format!("{}/p2p/{}", addr_str, peer_id));
+    }
+    manager.network = Some(net);
```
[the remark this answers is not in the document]
JonathanAldrich: This would be good to fix.  Can you address it?  The AIs suggest slightly different fixes, you could choose one of them or write your own.

JonathanAldrich (changes requested):
A few things identified by AI reviews to fix

meerkat-lib/src/runtime/manager/mod.rs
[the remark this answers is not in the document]
nivethadhanakoti: Agreed this is worth doing as a separate issue rather than rushing it in here. SendError only carries the Address in the UnreachableAddress case, so a complete fix needs msg_id-keyed eviction, capturing the MessageSent id when we emit an Update and tracking msg_id to listener. There is also a PeerDisconnected event that would let us drop all of a disconnected peer's subscriptions at once, which matches the shutdown cleanup you mentioned. Since both touch the send path, I would rather land them as a focused follow-up.

meerkat-lib/src/net/types.rs
```diff
@@ -139,6 +139,30 @@ pub enum MeerkatMessage {
     /// request is alive and still queued, so it keeps waiting
     /// instead of timing out
     WaitParked { request_id: u64 },
+
+    /// #24: subscribe `listener_service.listener_def` (living on the sender's
+    /// node) as a listener on `service.member` (on this node). The owner
+    /// registers it and replies with an initial `Update` carrying the current
+    /// value. Names are strings; symbols are interner-local.
```
andrewxzhou: I think this comment could be elaborated upon briefly. On its own "Names are strings; symbols are interner-local" states a fact that is true about the overall Meerkat codebase, but it doesn't really explain why that fact is relevant to the Update and RequestUpdate variant. I think you could elaborate this into something like "Since symbols only live in their respective nodes, while `MeerkatMessages` go between nodes, we represent services, members, etc, as Strings so that each node can understand the service the update pertains to"

meerkat-lib/src/runtime/interpreter/evaluator.rs
```diff
@@ -235,6 +235,18 @@ pub async fn eval(
             service_name,
             member_name,
         } => {
+            // #24: during a reactive recompute the cross-service deps are already
```
andrewxzhou: relatively minor, but there are a few grammatical errors hanging out in this comment, and it's not particularly clear. I would rewrite this as "during a reactive update, we check the cache first to see if we've already computed the dependencies for a service. If it's in the cache, we just use that instead of recomputing the dependencies with a `lookup`."

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -493,83 +570,290 @@ impl Manager {
...
+                self.send_oneway(Address::new(&reply_to), msg).await;
             }
         }
     }
 
+    /// #24 listener side: a remote member changed (or its initial value). Cache
+    /// it, recompute the dependent def from cache, and cascade to its listeners.
+    pub async fn handle_update(
+        &mut self,
+        listener_service: String,
```
JonathanAldrich: This method (and others in Manager) should take symbols rather than strings (all other methods are already written this way).  Clients should convert strings to symbols before calling into Manager.

meerkat/src/main.rs
```diff
@@ -533,6 +576,126 @@ async fn run_server(
     }
 }
 
+async fn run_watch(
```
JonathanAldrich: There's a lot of duplicated code here.  Please refactor to avoid duplicating any substantial code.  One possibility is to just put the unique code here (I think just the loop at the end?) into an if statement within run_client.  Other possibilities include moving duplicated code into functions that can be called from run_watch/run_client/run_server.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -20,6 +20,13 @@ pub struct Service {
     pub vars: HashMap<Symbol, VarState>,
     pub defs: HashMap<Symbol, Expr>, // original def expressions for re-evaluation
     pub dep: DependAnalysis,         // dependency graph + topo order
+    /// #24: who depends on each member: member -> {(listener service id, def)}.
+    pub listeners: HashMap<Symbol, HashSet<(ServiceNetId, Symbol)>>,
+    /// #24: cached values of each def's cross-service deps:
+    /// def -> {(source service, member) -> value}.
+    pub dep_cache: HashMap<Symbol, HashMap<(Symbol, Symbol), Value>>,
+    /// #24: each def's direct cross-service deps as (service, member) symbols.
+    pub dep_remote: HashMap<Symbol, HashSet<(Symbol, Symbol)>>,
```
JonathanAldrich: Do we need this in addition to `dep_cache`?  Isn't this just storing the keys from the HashMaps in `dep_cache`?

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -493,83 +570,290 @@ impl Manager {
...
             let event = n.try_recv_event();
             match event {
                 Some(NetworkEvent::MessageReceived { msg, .. }) => {
+                    // #24: these carry no request_id and are not replies; buffer
+                    // them for async draining instead of dropping them here.
+                    if matches!(
+                        msg,
+                        MeerkatMessage::RequestUpdates { .. } | MeerkatMessage::Update { .. }
+                    ) {
+                        self.pending_reactive.push(msg);
```
JonathanAldrich: Why put them in a queue and handle them later?  You could just handle them here, wouldn't that work just as well and be simpler?

meerkat-lib/src/runtime/manager/mod.rs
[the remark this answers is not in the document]
JonathanAldrich: @coderabbitai create a follow up GitHub issue for this

JonathanAldrich (changes requested):
Can you look at the design issues below please?  Thanks!

meerkat-lib/src/runtime/manager/mod.rs
andrewxzhou: Seconding this - from what I can tell you could just make `dep_cache` a `HashMap<Symbol, HashMap<(Symbol, Symbol), Option<Value>>>` and where you currently initialize `dep_remote`, initialize `dep_cache` to have all `None` values instead

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -296,13 +334,52 @@ impl Manager {
...
+
+            // #24: now that init succeeded, register listener edges so a change to
+            // a member notifies the defs that depend on it. Same-service deps come
+            // from dep_graph; cross-service deps come from dep_remote, where a local
+            // owner is wired in-process and a remote owner is subscribed over the
+            // wire. Only runs on the success path: a rolled-back service must not
+            // register listeners.
+            if let Some(this_id) = self.services.get(&svc_name).map(|s| s.id.clone()) {
+                let mut edges: Vec<(Symbol, Symbol, Symbol)> = Vec::new();
+                if let Some(s) = self.services.get(&svc_name) {
```
andrewxzhou: This bit seems redundant, since you also do this `self.services.get` in the outer if expression. It probably doesn't make that much of a difference, but it'd be nicer to move this step to the outer `if` and replace the `let Some(this_id) = ...` with `s.map(...)`

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -493,83 +570,290 @@ impl Manager {
     }
 
     async fn propagate(&mut self, service_name: Symbol, changed_var: Symbol) {
-        // collect defs that need re-evaluation in topo order
-        let topo_order: Vec<Symbol> = self
+        // #24: event-driven reactivity over the listener graph. A change to a
+        // member notifies its listeners; each local listener recomputes from
+        // current values (and cached cross-service deps) and, if it changes,
+        // cascades to its own listeners. A listener resolving to another node is
+        // notified over the wire.
```
andrewxzhou: i feel like this comment could be a little more precise? Specifically, it would help the clarity a bit if you noted somewhere the `Some` case corresponds to a local update which adds to the worklist, while the `None` case corresponds to a remote update which is handled by `emit_update`

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -493,83 +570,290 @@ impl Manager {
...
+            &expr,
+            &env,
+            &mut EvalContext {
+                manager: self,
+                service_name: svc,
+                txn: None,
+            },
+        )
+        .await;
+        self.reactive_cache = None;
```
andrewxzhou, line 639: A comment to clarify setting the `reactive_cache` back to None would be nice - I was a bit confused by it when I first looked at it

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -493,83 +570,290 @@ impl Manager {
...
+        let expr = match self
             .services
-            .get(&service_name)
-            .map(|s| s.dep.topo_order.clone())
+            .get(&svc)
+            .and_then(|s| s.defs.get(&def))
+            .cloned()
+        {
+            Some(e) => e,
+            None => return false,
```
andrewxzhou: It might be good practice to add a `log::warn` similar to line 653 here, since we shouldn't expect to not be able to find the def we're trying to recompute

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -493,83 +570,290 @@ impl Manager {
...
+            .services
+            .get_mut(&svc)
+            .and_then(|s| s.vars.get_mut(&def))
+        {
+            Some(var_state) => {
+                let differs = var_state.value != value;
+                var_state.value = value;
+                differs
+            }
+            None => false,
```
andrewxzhou: Similar to line R622, but doubly so - because this suggests that we found the def and recomputed it, but now it's not here anymore

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -493,83 +570,290 @@ impl Manager {
...
+    async fn emit_update(
+        &mut self,
+        listener_id: &ServiceNetId,
+        listener_def: Symbol,
+        svc: Symbol,
+        member: Symbol,
+    ) {
+        let reply_to = match self.listener_addrs.get(listener_id) {
+            Some(a) => a.clone(),
+            None => return,
```
andrewxzhou: This is also symmetric to R622 where I think a `log:;warn` would be nice. Ditto for all the other early returns throughout this function

meerkat-lib/src/runtime/semantic_analysis/var_analysis/read_write.rs
```diff
@@ -3,6 +3,73 @@ use crate::runtime::interner::Symbol;
...
 
 impl Expr {
+    /// Collect every cross-service reference `(service, member)` appearing
+    /// anywhere in this expression. Counterpart to `free_var`, which drops
+    /// `MemberAccess`; issue #24 needs these to know which remote members a
+    /// def must subscribe to. Symbols are interner-local, resolved to strings
+    /// only at the network boundary.
+    pub fn cross_service_deps(&self) -> HashSet<(Symbol, Symbol)> {
+        match self {
+            Expr::Literal { .. } | Expr::Table { .. } | Expr::Variable { .. } => HashSet::new(),
```
andrewxzhou: Shouldn't `Expr::Table` be pulled out into a separate match branch with something to denote that it is unimplemented?

meerkat/src/main.rs
```diff
@@ -515,6 +521,47 @@ async fn run_server(
...
+                    manager
+                        .handle_update(
+                            ServiceNetId(listener_service),
+                            listener_def_sym,
+                            source_sym,
+                            member_sym,
+                            value,
+                        )
+                        .await;
+                }
```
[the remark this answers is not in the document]
nivethadhanakoti: Good catch — the interning at the dispatch boundary happens before validation, so malformed remote names can grow the interner. The fix that fits the "symbols cross the Manager boundary" direction is a lookup-without-insert on the interner (resolve existing names, discard if absent), which needs a small interner API addition. Since it's a hardening change that touches the interner and sits next to the listener-eviction work, would you prefer I do it here or fold it into a follow-up alongside #108?

nivethadhanakoti:
@JonathanAldrich CodeRabbit flagged that RequestUpdates/Update names get interned at the dispatch boundary before they're validated, so malformed remote names could grow the interner. It's fine deferring to #108, but since the clean fix (a lookup that doesn't insert + validating before interning) sits a little against the "Manager takes symbols, clients convert first" direction, wanted your call: fix here or track in #108?

meerkat-lib/src/runtime/manager/mod.rs
nivethadhanakoti: For the queue: agreed that it wasn't necessary. I updated ⁠ dispatch_network_events ⁠ to be ⁠ async ⁠ and handle ⁠ RequestUpdates ⁠/⁠ Update ⁠ inline, so the intermediate buffer has been removed.

meerkat-lib/src/runtime/manager/mod.rs
nivethadhanakoti: That's a good point. Instead of changing `dep_cache` to store `Option` values, I removed `dep_remote` entirely and now derive the cross-service edges directly from each registered definition's `cross_service_deps()`. This keeps `dep_cache` values-only while making the registered definitions the single source of truth. I'm also happy to switch to the `Option` approach if you think that's a cleaner design.

meerkat-lib/src/runtime/semantic_analysis/var_analysis/read_write.rs
nivethadhanakoti: Thanks for the thorough review, @andrewxzhou! I addressed all of these in the latest push: expanded the wire-strings comment to explain why symbols can't cross nodes, rewrote the cache-check comment for clarity, removed the redundant `services.get`, clarified the `Some`/`None` propagation comment for the local vs. remote case, documented the `reactive_cache` reset, added `log::warn` on the early returns in `recompute_def` and `emit_update`, and yes, I split Expr::Table into its own match arm as suggested.

dustinjuliano:
It looks like the way this PR is currently using the `interner` is:

1. Receive a string name from the network
2. Call `interner.insert(name)` to get a `Symbol`
3. Look up that `Symbol` in `manager.services` or `svc.vars`
4. If absent, reject the message

However, an earlier PR #81 introduced both the `interner` and the `net::codec` , and `net::codec` was intended to be the sole source for encoding and decoding messages over the network. It guards the zero-trust boundary between the process (a Meerkat node) and information going over the wire.

It appears that new messages may have been added and their corresponding encoding and decoding logic, tests, and validation for zero-trust security were not added in `net::codec`. So, instead of routing requests through that security checkpoint, mutable state attempts were made to `interner`.

Additionally, the `interner` should be thought of as an append-only data structure by design. This is actually a safety mechanism that allowed catching this kind of unintended use. The only modules that should be permitted to append or write to `interner` are:

- `runtime::parser`
- `net::codec`

All other parts of the workspace should consider `interner` to be read-only/immutable. This is by design to protect the internal state and security of the process. It also makes maintenance easier, as we know there are only a handful of locations in the entire project that are permitted to alter `interner`.

(I propose opening a new issue to refine the access control using advanced Rust visibility features such as `in path`. This way it will be possible to use Rust to enforce read/write design decisions on `interner` to ensure it cannot be altered accidentally.)

The other issue related to `interner` is that it should not be used as a lookup mechanism for scoping or binding. While it is true currently that a `service` name is unique locally to a Node, consider what happens in this Meerkat pseudo-code fragment:

```
service foo {
	var bar = 1;
}

service bar {
}
```

If the `interner` is used as a lookup mechanism, it will result in a false positive. This is because `bar` has been mapped by the string `"bar"` from that `var` identifier, which is a member of `foo` and was parsed prior to `service bar`. However, if someone were to then try to query `interner` to see if `bar` exists, it would return a valid, interned `Symbol` or a `String` if doing a reverse lookup. If that information were then used that to indicate that the *service* `bar` already exists then that would be in error. Therefore, `interner` can never be used to make decisions on whether or not a member field is extant for a type (like a service).

As a result, `interner` cannot be used as a substitute for lexical scoping or binding. It should be noted that this was proposed in Issue #80 and is about to be introduced in an upcoming PR. I believe there may be extant functionality to check if a `service` exists inside `Manager`, which could be used in the interim in this PR until that new PR arrives with name resolution.

I have also identified some possible bugs, also related to the use of the `interner`, and for similar reasons related to the above example. It will take me some more time to provide those comments, as I want to be certain they are correct.

I am also working on a potential high-level overview for a fix, which I will post in a follow up comment.

dustinjuliano:
Here is the comment I promised describing some high level steps for a correction to the way the `interner` is being used. First, I would begin by changing this line in `interner.rs`, at least temporarily while making the fix, but it could become part of this PR since it is a tiny edit:

In `meerkat-lib/src/runtime/interner.rs`, change this line:

```
pub fn insert(&mut self, s: &str) -> Symbol
```

to

```
pub(crate) fn insert(&mut self, s: &str) -> Symbol
```

That will make it impossible to append to the `interner` instance from `main.rs`. This will restrict access until I can maybe come up with a more fine-grained solution using Rust module permissions.

Then I would recommend enhancing `net::codex` with support for the new messages added by this PR, including:

- encoding
- decoding
- validation
- associated tests

The `net::codec` (which I will just refer to as "codec" hereinafter) is designed to be the single entry point for network serialization and contains all the validation. The high-level steps for this are:

Within codec:

1. Add functions for encoding/decoding the newly added `RequestUpdates` and `Update` as with other message forms
2. Call `validate_identifier` prior to `interner.insert` to protect it; similar behavior is present in the other functions for the messages
3. No code outside codec should call `interner.insert` on data received over the wire

This would address numerous other bugs/issues I have located so far because that is the root cause. The codec needs to be extended for the newly added messages and no write calls to the `interner` used.

Recommended new functions in `net::codec`:

 - `decode_request_updates`
- `decode_update`

A program sketch as a guide:

```
validate_identifier(service)?;
validate_identifier(member)?;
validate_identifier(listener_def)?;
Ok((
	interner.insert(service),
	interner.insert(member),
	interner.insert(listener_def),
))
```

These transformations may be needed in multiple locations, apparently related to `Update` in `Manager::dispatch_network_events`, and to the `Update` arm of the watch loop in `run_client` (main.rs lines 716–718).

A small nitpick is the use of the variable name and/or prefix `member` , as it does not immediately self-identify as what it is referring to or its type. The receiver's service struct has two distinct entities: `service.vars` (for `VarDecl`) and `service.defs` (for `DefDecl`). Both are keyed by `Symbol` in the same `vars` map. "Member" is therefore an ambiguous and could be interpreted to refer to either at a glance.

The second part of the fix is to use the right authority in the codebase to query services. There appears to be a correct usage example in `handle_request_updates` to use as a starting point, as least conceptually:

```
let member_exists = self
    .services.get(&service_sym)
    .map(|s| s.vars.contains_key(&member_sym))
    .unwrap_or(false);
```

The `lookup` function (mod.rs line 396) receives a `Symbol` that came from trusted code and queries the service registry directly:

```
self.services.get(&service_name)
    .and_then(|s| s.defs.get(&var_name))
```

No interning happens inside `lookup`. The symbol is already validated by the time it arrives.

This part of the fix depends on the extension to `net::codex`, though.

I hope this was helpful. I see this is a large update. I was tasked with only reviewing this part involved with the `interner`, and this completes my review. If you have any questions please feel free to reach out by any means, Slack or otherwise.

meerkat/src/main.rs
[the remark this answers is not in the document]
dustinjuliano: Please see related comments on the intended use of `interner`. I do not recommend following the advice by Code Rabbit on this one. The comments:

- https://github.com/meerkat-lang/meerkat/pull/90#issuecomment-4826258350
- https://github.com/meerkat-lang/meerkat/pull/90#issuecomment-4826885764

dustinjuliano (changes requested):

ShreejithBabuG:
@dustinjuliano Thank you for the detailed review and for taking the time to explain the reasoning behind it. It was very helpful.

We weren't aware of the design invariants around the interner when implementing this feature, and since we hadn't followed the approach proposed in Issue #80, we ended up using the interner in a way that wasn't aligned with its intended purpose. Your explanation about keeping 'net::codec' as the sole entry point for network serialization and treating the interner as append-only made the architectural intent much clearer.

We'll work on restructuring the implementation accordingly by routing the new messages through 'net::codec', adding the necessary validation and tests, and using the appropriate service lookups instead of relying on the interner for existence checks.

Thanks again for the thorough review and the guidance.

meerkat-lib/src/net/codec.rs
```diff
@@ -32,6 +32,51 @@ fn validate_string_literal(s: &str) -> Result<()> {
     Ok(())
 }
 
+/// #24: validate and intern the identifier fields of a `RequestUpdates`
+/// message arriving over the wire. Per the zero-trust boundary, all interning
+/// of network-sourced names happens here in codec (never in the manager or
+/// main), after `validate_identifier`. `listener_service` and `reply_to` are
+/// network addresses, not identifiers, so they are not interned here.
```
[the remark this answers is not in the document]
nivethadhanakoti: Good point — validate_identifier was length-only, so non-identifier strings still passed. I've tightened it to match the lexer's identifier rule ([A-Za-z_][A-Za-z0-9_]*), so wire names the parser wouldn't accept locally are now rejected before interning, with a test. Since this strengthens the shared codec validator, it covers the other decode paths too.

meerkat-lib/src/net/codec.rs
```diff
@@ -32,6 +32,51 @@ fn validate_string_literal(s: &str) -> Result<()> {
...
+) -> Result<(Symbol, Symbol, Symbol)> {
+    validate_identifier(listener_def)?;
+    validate_identifier(source_service)?;
+    validate_identifier(member)?;
+    Ok((
+        interner.insert(listener_def),
+        interner.insert(source_service),
+        interner.insert(member),
+    ))
+}
```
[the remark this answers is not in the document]
nivethadhanakoti: Valid, and this is a real spoofable path — a peer can name another service's listener_service with its own reply_to and redirect that update stream. The correct fix binds the return route to the authenticated transport sender rather than trusting the payload, which needs transport-level identity. That's the same zero-trust boundary @dustinjuliano is building in #80/#111, so I'd like to coordinate it there rather than land a partial fix here. Tracking it for the hardening follow-up.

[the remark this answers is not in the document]
dustinjuliano: Would recommend making an issue to track this so that we do not miss it in a follow up round. This is because issue #80 is closed, however, so if we do not track this as a bug or enhancement it will be lost when the PR is closed. #111 is not related to this and is likely to be closed due to further investigation on the way Rust module-level access control works.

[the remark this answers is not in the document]
dustinjuliano: @coderabbitai Please make an issue for this based on context. The issues cited, such as #80 and #111 are closed and when this PR is closed we will lose the track of this. It should be classified as a (security) bug. Please do not make a duplicate if you can avoid it and be as concise as possible, including the code fragment and your diagnosis.

JonathanAldrich:

Issue #92, opened 2026-06-23
Panic on Division by Zero


dustinjuliano:
Reported from Code Rabbit while doing PR #91  The interpreter may panic when performing a division by zero.

I began looking at the fix and found right away that it may require modifications to the error enum, which put it beyond the scope of PR #91, which is already moderate. I would propose introducing a kind of runtime error or evaluator error to distinguish from a type error.

PR #103, opened 2026-06-26
fix: prevent arithmetic panics in interpreter

dustinjuliano:
This PR addresses Issue #92 and expands on that original report to address known panics for built-in Rust arithmetic in the Meerkat interpreter. The only operation not covered is modulus, and a comment is place in the source to help hint that it should also be checked using similar logic as this fix.

A sample Rust panic demo program was constructed to confirm and also illustrate:

```rust
fn main() {
  let zero = std::hint::black_box(0);
  let minus_one = std::hint::black_box(-1);
  let _ = 1 / zero;
  let _ = 1 % zero;
  let _ = i32::MIN / minus_one;
  let _ = i32::MIN % minus_one;
}
```

Each of those is a separate panic issue. These happen at runtime. More panics happen in debug builds than release builds, and this PR narrowly tailored to just release build panics, as it could potentially terminate a Meerkat Node.

This fix is implemented by introducing checked division and is an isolated change. Tests added to help validate. An `RuntimeError` was added to the error enum for future expansion and to properly categorize this kind of error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Division errors now return clear runtime error messages instead of crashing in edge cases like division by zero or integer overflow.
  * Error handling was broadened so more runtime error cases are treated consistently during execution.
  
* **Tests**
  * Added coverage for division-by-zero and overflow behavior to verify errors are reported safely and consistently.

### DIFF
--- a/meerkat-lib/src/runtime/interpreter/evaluator.rs
+++ b/meerkat-lib/src/runtime/interpreter/evaluator.rs
@@ -15,6 +15,7 @@ pub enum EvalError {
     WaitDieAbort(String),
     WaitOn(Symbol, Symbol),
     AssertionError(String),
+    RuntimeError(String),
 }
 
 /// Implement the `Display` trait for the `EvalError` type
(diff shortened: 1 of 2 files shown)


dustinjuliano:
@coderabbitai full review

dustinjuliano:
Ready for human review.

JonathanAldrich (approved):
Looks good!

PR #127, opened 2026-07-04
code update design thoughts

JonathanAldrich:
## Summary by CodeRabbit

* **Documentation**
  * Added a new design document outlining how code updates should be expressed, including a clearer distinction between modifying existing definitions and introducing new ones.
  * Described a “service class” approach so shared service changes can propagate to instantiated services.
  * Documented dependency-aware type checking/compatibility validation and an update coordination strategy using write-locking, with defined behavior for concurrent update conflicts and type errors.

### DIFF
--- a/docs/design/code-updates.md
+++ b/docs/design/code-updates.md
@@ -3,8 +3,11 @@
 ## Syntax
 
 The syntax for code updates should distinguish updating existing definitions
-from introducing new definitions.  One possibility would look like this
-(example borrowed from the Requirements wiki page):
+from introducing new definitions.  This expresses the intent to update and
+ensures that the developer does not update a definition by accident when they
+think they are adding a new definition.  The `update` keyword indicates
+services and members that are updated.  One possible syntax would look like
+this (example borrowed from the Requirements wiki page):
 
 ```meerkat
 update s1 {
@@ -29,19 +32,40 @@ service class MyServiceClass {
 service myService = MyServiceClass;
 ```meerkat
 
+When a service class is updated, then all the services that are an instance
+of that service class are updated.
+
+The existing `service` declaration can be thought of as declaring a service
+class and instantiating it to a `service` that has the same name.  In
+contrast, a `service class` is not instantiated until it is done explicitly
+with a separate `service` declaration like the one described above.
+
 ## Typechecking
 
 A starting point for type checking code updates is found in
 [Domingues and Seco 2015](https://docentes.fct.unl.pt/jrcs/files/rebls15.pdf).
 The type checker knows about all the dependencies of a service, so if a
-code update changes the interface of a service, the type checker will verify
+code update changes the interface of a service in a non-monotonic way (e.g.
+by removing a member, or by changing to an incompatible type, such as from
+`int` to `string`), the type checker will verify
 that client code is not broken (and/or an update to the client code is
 packaged with the service code update).
 
 It should be possible to type check a code update against code in one or more
 `.mkt` files, or against one or more running services.  In the latter case,
 the code update is guaranteed to succeed unless the code is changed by some
-other update in between.
+other update in between.  This is a version of the usual type
+soundness property, extended to cover code updates.
+
+Note that to make this work in practice, when a variable or table's type is
+changed, migration code must be written (as shown in Domingues and Seco) in
+order to populate the value of the new variable/table to conform to the
+new type.  Typically this migration code reads from the old variable/table
+and writes to the new one, but it can read from other sources as well if
+need be.  The migration code is executed synchronously with the code update,
+and once the code update is done, the migration code is no longer needed
+as all the data has been migrated.  It can be kept around for replay
+purposes, if desired.
 
 ## Coordination
 
@@ -54,6 +78,13 @@ on those imported services.
 
 When Meerkat supports an import statement in the REPL accepting a URL
 (Issue #95), then we can easily import any service we want to update.
+For now, import by URL means importing a service that is running on some other
+Meerkat node and is exposed through that URL.
+
+
+Eventually, code updates will likely be limited to code that is defined in the
+current Meerkat distributed virtual machine (DVM), for security reasons.
+Meerkat DVMs may of course choose to extend some level of trust to each other.
 
 When an update is executed, we need to acquire a write-lock on any service
 being deleted and on any members being deleted or modified.  That write-lock
--- a/docs/design/code-updates.md
+++ b/docs/design/code-updates.md
@@ -62,7 +62,8 @@ changed, migration code must be written (as shown in Domingues and Seco) in
 order to populate the value of the new variable/table to conform to the
 new type.  Typically this migration code reads from the old variable/table
 and writes to the new one, but it can read from other sources as well if
-need be.  The migration code is executed synchronously with the code update,
+need be.  The migration code is executed as part of the code update
+transaction, while code update locks are held,
 and once the code update is done, the migration code is no longer needed
 as all the data has been migrated.  It can be kept around for replay
 purposes, if desired.
@@ -85,6 +86,7 @@ Meerkat node and is exposed through that URL.
 Eventually, code updates will likely be limited to code that is defined in the
 current Meerkat distributed virtual machine (DVM), for security reasons.
 Meerkat DVMs may of course choose to extend some level of trust to each other.
+We defer the design of this authorization boundary to when we design the DVM.
 
 When an update is executed, we need to acquire a write-lock on any service
 being deleted and on any members being deleted or modified.  That write-lock
@@ -95,7 +97,9 @@ canceled, and error messages are reported to the developer.
 
 ## Errors due to code update races
 
-A code update will always preserve the code in a working state.  However, there
+A code update is transactional; if it succeeds, it will always preserve
+the code in a working state, and if it fails, all changes are rolled back.
+However, there
 could be races between executing some action and code updates.  Whichever
 acquires the appropriate locks will run.  If the code update runs first, some
 triggered actions could be invalidated.  We need to design a mechanism to report
--- a/docs/design/code-updates.md
+++ b/docs/design/code-updates.md
@@ -6,11 +6,11 @@ The syntax for code updates should distinguish updating existing definitions
 from introducing new definitions.  One possibility would look like this
 (example borrowed from the Requirements wiki page):
 
-```
+```meerkat
 update s1 {
   update def y = x + 5;
 }
-```
+```meerkat
 
 ## Service classes
 
@@ -21,13 +21,13 @@ service class.  A service class is a template for instantiating services; the
 most common use case is web clients, but there could be others.  Possible
 syntax for service class definition and instantiation:
 
-```
+```meerkat
 service class MyServiceClass {
   ...
 }
 
 service myService = MyServiceClass;
-```
+```meerkat
 
 ## Typechecking
 
@@ -52,14 +52,14 @@ REPL at that node, `update` statements can be executed to update them.  If the
 node imports services from other nodes, it can also execute `update` statements
 on those imported services.
 
-When Meerkat supports a import statement in the REPL accepting a URL
+When Meerkat supports an import statement in the REPL accepting a URL
 (Issue #95), then we can easily import any service we want to update.
 
 When an update is executed, we need to acquire a write-lock on any service
 being deleted and on any members being deleted or modified.  That write-lock
 will prevent any transactions from reading or writing to the locked members.
 The type checker is run after locks are acquired but before applying the code
-update, and if there are type errors, the locks a released, the update is
+update, and if there are type errors, the locks are released, the update is
 canceled, and error messages are reported to the developer.
 
 ## Errors due to code update races
@@ -67,6 +67,6 @@ canceled, and error messages are reported to the developer.
 A code update will always preserve the code in a working state.  However, there
 could be races between executing some action and code updates.  Whichever
 acquires the appropriate locks will run.  If the code update runs first, some
-triggered actions could be invaliated.  We need to design a mechanism to report
+triggered actions could be invalidated.  We need to design a mechanism to report
 an error to the user in such cases, and/or for the programmer to catch any
 errors that are triggered by the update and run some other compensating action.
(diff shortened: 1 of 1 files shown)


docs/design/code-updates.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Syntax
+
+The syntax for code updates should distinguish updating existing definitions
+from introducing new definitions.  One possibility would look like this
+(example borrowed from the Requirements wiki page):
+
+```meerkat
+update s1 {
+  update def y = x + 5;
```
dustinjuliano, line 14: Interesting. Thinking of how this might be implemented as well. It makes me imagine a structural comparison between `s1` A and `s1` B. Having an explicit `update` keyword shows intent. Could also infer `update` on the body of the updated service. Some thoughts:

- An empty update block/definition means no changes; the no-op of an update
- If a binding is present, no `update` keyword explicitly required as the system can perform a structural comparison, and if the bodies differ, it updates the body
- Alternatively, require having the updated `s1` provide the complete and full definition, and those bindings that are not marked `update` will be preserved
  - This raises question, though, of what happens if the provided `s1` bindings do not match the existing one. Could perform a check, raise an error, etc

docs/design/code-updates.md
```diff
@@ -0,0 +1,72 @@
...
+```meerkat
+
+## Service classes
+
+Web clients will probably want to instantiate services defined on the server
+that they connect to.  When we update the definition of the service we'd like
+all instances of that service to be updated.  That suggests the idea of a
+service class.  A service class is a template for instantiating services; the
+most common use case is web clients, but there could be others.  Possible
+syntax for service class definition and instantiation:
```
dustinjuliano, line 25: I would have to ask about this more, as my interpretation is that the intent is for `service` to become class-like, as this is something you told me early on. Since it was at the very beginning of the project, I did not have as much understanding of Meerkat internals as I do now, so I have some potentially helpful thoughts to share.

My present design goal for the type system, which I am updating as I learn more and gain more information, is to treat each `service` (represented by `Stmt::Service`) as both definition and declaration.

Perhaps this looks like introducing a `ServiceClass` (or some other name, just choosing this for discussion) in `runtime::tt::types.rs` and a corresponding updates to `net::codec` for transport.

- My understanding is that `ServiceClass` would *not* be a first-class value
- `ServiceClass` and infrastructure would treat it as *eventually mutable* for the purpose of live code updates
  - Stated to contrast with marking it and/or designing the internals around types that never change
  - Having the ability to update a type definition has wide-ranging architectural impacts, so stating it for clarity

So, a `service` can be type checked and this could construct the `ServiceClass` or `ServiceType`, which would be used to for inference and type checking. This information would *persist* in the environment system (which is being introduced in PR #120), so that it would be not only available for live code updates but for any other purposes, such as dependency analysis, and so forth.

I would then ask how the above mental model I have could be revised based on what you had in mind.

docs/design/code-updates.md
JonathanAldrich: Yes, intent is exactly why I was thinking it would be useful to require an `update` keyword. That way one doesn't do an update accidentally.  I was indeed thinking that anything not mentioned would be preserved.

I was assuming it is OK if the newly provided bindings do not match the existing ones, as long as the code type checks after all updates are applied.  This is how REBLS'15 works, it allows non-monotonic updates which adds expressivity.

docs/design/code-updates.md
JonathanAldrich: My mental model is that `service` declares a class and instantiates it.  So it is both like a class and like an object.  The `object` keyword in Scala is similar.  The point of having a `service class` is (A) other code can use it to instantiate--although I suppose we could do that for regular `service`s as well; and (B) unlike `service` it is not automatically instantiated.  So it's good for a client UI which is declared on the server but not instantiated (part B), but needs to be instantiated on multiple clients (part A).

docs/design/code-updates.md
```diff
@@ -0,0 +1,72 @@
...
+[Domingues and Seco 2015](https://docentes.fct.unl.pt/jrcs/files/rebls15.pdf).
+The type checker knows about all the dependencies of a service, so if a
+code update changes the interface of a service, the type checker will verify
+that client code is not broken (and/or an update to the client code is
+packaged with the service code update).
+
+It should be possible to type check a code update against code in one or more
+`.mkt` files, or against one or more running services.  In the latter case,
+the code update is guaranteed to succeed unless the code is changed by some
+other update in between.
```
dustinjuliano: I peeked (very briefly only) at BEAM in Erlang for hot code updates, because I was trying to see how other systems had solved it. Important points of comparison:

- Dynamic versus static type systems
- Updates to the schema or shape of types versus narrow updates to existing bindings

In all cases, I got caught on the "guaranteed" part of the paragraph. I would need to ask for clarification.

docs/design/code-updates.md
```diff
@@ -0,0 +1,72 @@
...
+
+A design choice we might consider is how to manage code updates in a Meerkat
+distributed virtual machine.  First, a simple approach for specifying updates.
+Services and service classes live where they were originally defined.  In the
+REPL at that node, `update` statements can be executed to update them.  If the
+node imports services from other nodes, it can also execute `update` statements
+on those imported services.
+
+When Meerkat supports an import statement in the REPL accepting a URL
+(Issue #95), then we can easily import any service we want to update.
```
dustinjuliano, line 81: Is the intent to allow imports to be agnostic of origin? That is to say, in this proposed enhancement, an `import` would essentially be accessible to the full scope of any URI, which implies web, file, or any resource supported by that?

docs/design/code-updates.md
JonathanAldrich: I said that because the main (only?) reason a code update could fail is if it breaks typing.  So if you type check ahead of time, and then you don't change anything, the code update will succeed.  It's basically a type soundness guarantee extended to code updates.

docs/design/code-updates.md
```diff
@@ -0,0 +1,72 @@
...
+
+When an update is executed, we need to acquire a write-lock on any service
+being deleted and on any members being deleted or modified.  That write-lock
```
dustinjuliano, line 92: In my studies on these kinds of systems, deletion is one of the more difficult challenges, and I would have to ask about implementation, versioning, and consistency model.

docs/design/code-updates.md
JonathanAldrich: For now I am assuming that an import in the REPL by URL means an import of some service that is running on another Meerkat node.  So limited in scope.  And when we define the boundaries of a Meerkat distributed virtual machine, code updates will probably by default be limited to code in the current virtual machine.

dustinjuliano (commented):
I have left comments on specific sections. Here are my general thoughts:

## Load/Store

What if we treated the in-memory representation and state of a Meerkat program like load/store in a processor. The motivation behind this is to abstract the source of any updates. So, it could be from the REPL, network, or file system, and, from the perspective of whatever system is doing the live code update logic, it becomes origin agnostic.

## Purely Functional Inspiration

I had raised the idea of taking a bit of inspiration from [Datomic](https://docs.datomic.com/datomic-overview.html) to handle the distributed nature of Meerkat in the presence of updates. That website has very good documentation and could be helpful for ideas. The idea I latched onto the most was the purely functional nature of the database, which helps address numerous challenges, including consistency.

## Schema Invalidation and Data Migration

I also did a brief look into some of the fundamental challenges of updating a schema or type or some live instance of data. Meerkat has this in at least two ways it seems:

1. (Future/Tentative) `ServiceType` or `ServiceClass`
2. `TableType`

In both of these situations, extant instances of those types would potentially be invalidated or enter an inconsistent/unknown state. I found that there is some kind of principled mathematical structure which I think is a "migration function", but I am not yet familiar. This is essentially a function $F: A \to B$, where $A$ is the old service or table type and $B$ is the new one, and $F$ is the function that migrates all extant instances to the new type/schema. I raise this because it appears there is existing work on this (have not yet had time to go into it and you would know way more about it than me) and it appears to be some kind of canonical shape to the problem space.

Essentially, something has to find $F$, either automatically or it is supplied by the programmer. From what I saw, only briefly, it is very, very hard to synthesize $F$ automatically by just looking at the delta of one service and another, and this is because it is not merely a matter of structure, but semantics. That last part surprised me when I was looking into it. For very simple changes to services (or table types), like changing a variable in the body of a function, it might be easy to find $F$ automatically, but, in general, it might even be undecidable in the general case.

How this might affect Meerkat is that we might have to ask programmers to provide these transition functions $F$ inside or in addition to the `update` blocks/annotations for any part of the program they are changing. We could potentially infer what we can and then raise an error if Meerkat cannot automatically derive the migration, or cannot otherwise prove there would be no conflict.

These are just some thoughts, though, and only a very preliminary review of the related concerns, as I have been in full focus on solving PR #120. Just thought I would share to help start the discussion on those aspects. I hope this was helpful.

There is also some design that may be highly relevant to code updates in the upcoming PR #120, as I had to think about how environments could support live code update at the level of granularity you wanted (service and service field level). I believe those were addressed by the system and I am in-progress of finalizing it, testing it, and then lifting it out of my personal test/sandbox repo and into the feature branch for the PR.

docs/design/code-updates.md
dustinjuliano: Ah yes. The only counter to that I can think of is that the static checks for a potential merge/update is more general than type checking, and that type checking would only be one step. I mention in other comments about the migration function concept for invalidated extant data that enters an inconsistent state when a service changes.

If we think of the data as a term inhabiting a type, and that type changes, then its term is no longer necessarily an inhabitant of that type anymore. The latter comment about the migration function `F: A -> B` is a kind of process/program saying give me a proof of `A` (the old service) and I will give you a proof of `B` (the new service being updated). And that finding `F` automatically is very hard, potentially undecidable, and that we might have to ask Meerkat programmers doing non-trivial `update` to give us `F` alongside any `update` that cannot be mechanically deduced.

I could be totally wrong above the above, though, as it was only a fast glance due to trying to focus on the environment system.

docs/design/code-updates.md
JonathanAldrich: Sure.  Look at how REBLS'15 handles this issue.  Basically the programmer indeed has to provide the transition functions.  Safety requires that all the updates and transition functions fit together.

I think deletion isn't that hard if you support other kinds of non-monotonic updates, as REBLS'15 does.

dustinjuliano:
Another thought is that if we do have programmers provide these "migration functions", then how long would they remain?

I will also look into REBLS`15.

JonathanAldrich:
The idea with migration is that all the data is migrated synchronously with the code update.  So the migration function is only needed right at that moment, it can be forgotten after the update is done.  Of course, for software engineering reasons (e.g. to undo or replay an update) one might want to keep it around--that's more a version control thing than an update thing, but we may want to think about it at some point.

dustinjuliano:
> The idea with migration is that all the data is migrated synchronously with the code update. So the migration function is only needed right at that moment, it can be forgotten after the update is done.

A couple questions on that:

1.  I would ask how this works in the presence of a delay or messages/requests in transit.
2.  What is meant by synchronous in this context? I interpret that to mean batch, all-at-once, and/or transactional; the migration function is fully applied to all data on the node right then and there? This seems to be affected by my question (1), especially if the migration function does not persist.

Expanding on (2): What if types and/or data are distributed in some sense and this goal of it being synchronous would require some kind of sequence point on the entire swarm or a consensus of some kind? This is probably a bad question, as my understanding on that aspect of the protocol and data model is still expanding. However, I wanted to risk asking it because it may be important for the notion of the migration function and/or not persisting old types or versions of schemas.

Also saw you made some changes to the document and will take a look.

JonathanAldrich:
> 1. I would ask how this works in the presence of a delay or messages/requests in transit.

The migration function is applied during code update, during which time all the code impacted by the update is locked.  So delays can affect the responsiveness of the system because locks are held for longer, but they cannot cause inconsistencies.

> 2. What is meant by synchronous in this context? I interpret that to mean batch, all-at-once, and/or transactional; the migration function is fully applied to all data on the node right then and there? This seems to be affected by my question (1), especially if the migration function does not persist.

Transactional.

> Expanding on (2): What if types and/or data are distributed in some sense and this goal of it being synchronous would require some kind of sequence point on the entire swarm or a consensus of some kind? This is probably a bad question, as my understanding on that aspect of the protocol and data model is still expanding. However, I wanted to risk asking it because it may be important for the notion of the migration function and/or not persisting old types or versions of schemas.

Yes, this is what the locks are about.  It is likely that code updates will require locks across a wide swath of the system.  It is not clear to me that we need code updates to be completely global, but the blast radius is large enough that many code updates may be essentially global in practice even if not in theory.

dustinjuliano:
> It is likely that code updates will require locks across a wide swath of the system. It is not clear to me that we need code updates to be completely global, but the blast radius is large enough that many code updates may be essentially global in practice even if not in theory.

Thank you, that helps my understanding a lot. I am thinking that there may be a useful trade-off between a (semi-)persistent kind of system that persists and may avoid or minimize locks by making the system essentially one distributed purely functional data structure. I don't know if it would be enough to make it lock-free, but it might be able to reduce locks or make them more fine-grained and improve throughput. The trade-off, I think, might be trading for an increase in space for the histories/journal.

My ideas here could be wrong, so I am offering it with caution. These are just sketches of ideas at this stage, and I have not fully had the opportunity to sit down and do a detailed plan or model of these parts of the system.

JonathanAldrich:
I have been thinking of code updates in database terms: mutations that are controlled with ACID transactions.  A functional programming model is certainly possible, and would enable certain features--e.g. the ability to run the old version of a service and the new version in parallel for a while.  Now that kind of parallel execution could be desirable by users (who can be reluctant to upgrade) but I am guessing that most service providers do not want that model, except perhaps if it is the only way to avoid an interruption in service.  The problem is (A) keeping the state in the old and new versions synchronized--something that is hard and we may not be able to support well in an automated way and (B) security issues or other business concerns that drive service providers to force users to upgrade.

Functional updates could also be an implementation technique for changes that are still atomic from the user's perspective.  But note that this is a leaky abstraction: when the change is "pushed" to the end client, that is a kind of mutation!  It just becomes a mutation from the app endpoint pointing at one functional implementation to pointing at another.

JonathanAldrich:
And yes, there are probably performance tradeoffs here, I haven't thought of them in depth.  Certainly maintaining multiple versions (of an entire database!) at least uses more space.  But of course the current proposal has migration functions, and the source and target databases will likely have to be live simultaneously anyway to run them, for the duration of the migration.

Issue #39, opened 2026-05-30
web client


JonathanAldrich:
Implement the [web client design](https://github.com/meerkat-lang/meerkat/wiki/Requirements) and get the example `counter` and `webClient` code from that page working.

JonathanAldrich:
This should be implemented after #24

JonathanAldrich:
@ShreejithBabuG's [ping pong prototype](https://github.com/ShreejithBabuG/ping_pong_libp2p) can serve as a guide for how to integrate WASM and libp2p networking

PR #128, opened 2026-07-05
PR 2 of 3 toward #39 (web client): server responds to requests for service code

ShreejithBabuG:
Second of three PRs for the web client. This adds the mechanism for a client to request a `.mkt` file from a server by path, so it can load and run that code locally, per the Requirements page. Also folds in the html AST-printing cleanup requested on the #123 merge.

## What this adds
- Three wire messages (`ServiceCodeRequest` / `ServiceCodeResponse` / `ServiceCodeError`). They carry only strings, so serde handles them and no runtime `Value`/`Expr` codec change was needed.
- Server: `run_server` answers a `ServiceCodeRequest` by returning the requested `.mkt` file's source (via `codec::build_service_code_response`, which also validates the length-bounded fields), or an error.
- Client: `Manager::fetch_service_source` sends the request and awaits the reply over the existing request/reply path. Processing the returned source through the normal program-loading path (services and imports) is done by the caller; the browser entry point in PR 3 wires this in.
- The request identifies a file by path (like a URL path), and the server returns the whole file so any imports and other services it needs come with it. We send raw source text rather than serialized decls, keeping this PR small. Server-side parsing and type checking is noted as future work (needed for early error catching and code updates).

## HTML print cleanup (from #123 review)
Removed `HtmlTemplate::parts()` and `HtmlPartView`, which leaked the template's text/expr shape. Added `HtmlTemplate::print_ast()`, called from the printer, so the html module owns its own AST printing and the representation stays hidden.

## Testing
A mock-network round-trip test (client requests a path, server returns the whole file source) and a validation error-path test (oversized path is rejected with `ServiceCodeError`). Both call the same `codec::build_service_code_response` the server uses, so they exercise the real handler.

## Out of scope
The wasm browser client, HTML host page, JS wrapper, and HTTP server (PR 3). The `run_server` glue is exercised via the manual client/server flow rather than an automated test.

## Summary by CodeRabbit

* **New Features**
  * Added a network “service code” request/response flow to fetch `.mkt` source by path, including success (source) and failure (error) replies.
  * Added a client API to fetch remote service source and updated the server to handle requests.
  * Introduced a configurable request-string length limit (4096).

* **Bug Fixes**
  * Added strict validation for requested path and reply address, including safe path resolution to prevent escaping the served directory.

* **Refactor**
  * Updated HTML template inspection: removed the public parts view/iterator and added an AST-based printing method.

* **Tests**
  * Added integration tests for request roundtrips and oversized path rejection, plus unit coverage for file resolution rules.

### DIFF
--- a/meerkat-lib/src/net/types.rs
+++ b/meerkat-lib/src/net/types.rs
@@ -167,23 +167,23 @@ pub enum MeerkatMessage {
         value: NetValue,
     },
 
-    /// #39: request the source code of a service class by name. A (browser)
-    /// client sends this to fetch a service it imports so it can parse and
-    /// instantiate the service locally. The service name is carried as a
-    /// string and resolved through the receiver's own interner, like the
-    /// other cross-node messages.
+    /// #39: request the source of a `.mkt` file by path. A (browser) client
+    /// sends this to fetch a file it imports so it can process it locally
+    /// (creating services and resolving further imports), like loading a
+    /// program normally. The path is a resource path, similar to a URL path.
     ServiceCodeRequest {
         request_id: u64,
-        service: String,
+        path: String,
         reply_to: String,
     },
 
-    /// #39: response to `ServiceCodeRequest` carrying the service's source
-    /// text. The requester parses it through its own parser (interning
-    /// locally) and instantiates the service.
+    /// #39: response to `ServiceCodeRequest` carrying the whole `.mkt` file
+    /// source. The requester processes it through the normal program-loading
+    /// path. The source is unbounded in length; only the request path and
+    /// reply address are length-limited.
     ServiceCodeResponse {
         request_id: u64,
-        service: String,
+        path: String,
         source: String,
     },
 
--- a/meerkat-lib/src/net/types.rs
+++ b/meerkat-lib/src/net/types.rs
@@ -166,6 +166,29 @@ pub enum MeerkatMessage {
         member: String,
         value: NetValue,
     },
+
+    /// #39: request the source code of a service class by name. A (browser)
+    /// client sends this to fetch a service it imports so it can parse and
+    /// instantiate the service locally. The service name is carried as a
+    /// string and resolved through the receiver's own interner, like the
+    /// other cross-node messages.
+    ServiceCodeRequest {
+        request_id: u64,
+        service: String,
+        reply_to: String,
+    },
+
+    /// #39: response to `ServiceCodeRequest` carrying the service's source
+    /// text. The requester parses it through its own parser (interning
+    /// locally) and instantiates the service.
+    ServiceCodeResponse {
+        request_id: u64,
+        service: String,
+        source: String,
+    },
+
+    /// #39: the requested service class was not found on this server.
+    ServiceCodeError { request_id: u64, error: String },
 }
 
 /// Errors that can occur when sending messages
--- a/meerkat-lib/src/net/types.rs
+++ b/meerkat-lib/src/net/types.rs
@@ -187,7 +187,8 @@ pub enum MeerkatMessage {
         source: String,
     },
 
-    /// #39: the requested service class was not found on this server.
+    /// #39: error responding to a `ServiceCodeRequest`, e.g. an invalid
+    /// request (oversized path/reply_to) or a missing resource.
     ServiceCodeError { request_id: u64, error: String },
 }
 
--- a/meerkat-lib/src/runtime/manager/mod.rs
+++ b/meerkat-lib/src/runtime/manager/mod.rs
@@ -1108,21 +1108,6 @@ impl Manager {
             .unwrap_or_else(|_| "127.0.0.1".to_string())
     }
 
-    /// Perform a remote variable lookup over the network
-    ///
-    /// Sends a lookup query to the node owning the remote service and registers
-    /// the local node as a transaction participant.
-    ///
-    /// Args:
-    ///     service (Symbol): The remote service symbol
-    ///     member (Symbol): The member/variable symbol within the service
-    ///     txn (Option<&mut Transaction>): The active transaction context
-    ///
-    /// Returns:
-    ///     Result<Value, EvalError>: The retrieved value, or a network/timeout error
-    ///
-    /// Raises:
-    ///     EvalError::LocalDispatchFailed: If a timeout or dispatch error occurs
     /// #39: Fetch the source of a `.mkt` file from a remote server by path.
     ///
     /// Sends a `ServiceCodeRequest` and awaits the reply, reusing the same
@@ -1169,6 +1154,21 @@ impl Manager {
         }
     }
 
+    /// Perform a remote variable lookup over the network
+    ///
+    /// Sends a lookup query to the node owning the remote service and registers
+    /// the local node as a transaction participant.
+    ///
+    /// Args:
+    ///     service (Symbol): The remote service symbol
+    ///     member (Symbol): The member/variable symbol within the service
+    ///     txn (Option<&mut Transaction>): The active transaction context
+    ///
+    /// Returns:
+    ///     Result<Value, EvalError>: The retrieved value, or a network/timeout error
+    ///
+    /// Raises:
+    ///     EvalError::LocalDispatchFailed: If a timeout or dispatch error occurs
     pub async fn remote_lookup(
         &mut self,
         service: Symbol,
--- a/meerkat-lib/src/runtime/manager/mod.rs
+++ b/meerkat-lib/src/runtime/manager/mod.rs
@@ -930,6 +930,13 @@ impl Manager {
                             MeerkatMessage::CommitResponse { request_id, .. } => Some(*request_id),
                             MeerkatMessage::AbortResponse { request_id, .. } => Some(*request_id),
                             MeerkatMessage::WaitParked { request_id, .. } => Some(*request_id),
+                            // #39: code responses are replies routed to the waiting client.
+                            MeerkatMessage::ServiceCodeResponse { request_id, .. } => {
+                                Some(*request_id)
+                            }
+                            MeerkatMessage::ServiceCodeError { request_id, .. } => {
+                                Some(*request_id)
+                            }
                             MeerkatMessage::Ping { .. }
                             | MeerkatMessage::Pong { .. }
                             | MeerkatMessage::Announce { .. }
@@ -940,6 +947,8 @@ impl Manager {
                             | MeerkatMessage::Commit { .. }
                             | MeerkatMessage::Abort { .. }
                             | MeerkatMessage::RequestUpdates { .. }
+                            // #39: an incoming code request is handled server-side, not a reply.
+                            | MeerkatMessage::ServiceCodeRequest { .. }
                             | MeerkatMessage::Update { .. } => None,
                         };
                         if let Some(request_id) = rid {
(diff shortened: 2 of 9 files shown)


meerkat-lib/src/net/types.rs
```diff
@@ -166,6 +166,29 @@ pub enum MeerkatMessage {
...
     },
+
+    /// #39: request the source code of a service class by name. A (browser)
+    /// client sends this to fetch a service it imports so it can parse and
+    /// instantiate the service locally. The service name is carried as a
+    /// string and resolved through the receiver's own interner, like the
+    /// other cross-node messages.
+    ServiceCodeRequest {
+        request_id: u64,
+        service: String,
```
JonathanAldrich: Since we're doing validation for other messages to the server, the service and reply_to strings should be limited in length.

meerkat-lib/src/net/types.rs
```diff
@@ -166,6 +166,29 @@ pub enum MeerkatMessage {
...
+        reply_to: String,
+    },
+
+    /// #39: response to `ServiceCodeRequest` carrying the service's source
+    /// text. The requester parses it through its own parser (interning
+    /// locally) and instantiates the service.
+    ServiceCodeResponse {
+        request_id: u64,
+        service: String,
+        source: String,
```
JonathanAldrich, line 187: source code can be unlimited in length though

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -1114,6 +1123,71 @@ impl Manager {
...
+    /// sanctioned interner-writer) and create each service it defines.
+    pub async fn fetch_and_instantiate_service(
+        &mut self,
+        service_name: &str,
+        server_addr: Address,
+    ) -> Result<(), EvalError> {
+        let source = self.fetch_service_source(service_name, server_addr).await?;
+        let stmts = crate::runtime::parser::parse_string(&source, &mut self.interner)
+            .map_err(EvalError::RemoteDispatchFailed)?;
+        for stmt in &stmts {
```
JonathanAldrich: This is executing the service code.  But the code might have statements other than a Service.  Most importantly it might have imports--because a service on the client almost certainly needs to talk with services on the server it came from.  I think it might be better to call existing functionality elsewhere that would already handle service creation and imports.  This would also avoid duplicate code.

meerkat-lib/src/runtime/parser/mod.rs
```diff
@@ -201,6 +201,51 @@ fn parse_expr_fragment(src: &str, interner: &mut Interner) -> Result<crate::ast:
...
+/// #39: Extract the source text of a single `service <name> { ... }` block
+/// from full program source, by locating the service header and brace-matching
+/// to its closing brace. Used server-side to answer a request for a service's
+/// code by name.
+///
+/// Returns `None` if no service with that name is found. Brace matching does
+/// not account for a `}` inside a string literal (the same limitation noted
+/// for the html lexer); source that already parsed cleanly is well-formed for
+/// this purpose.
+pub fn extract_service_source(source: &str, service_name: &str) -> Option<String> {
```
JonathanAldrich: I think we don't need to do this--we should just return an entire `.mkt` file. That way if multiple services are needed they can all be included in the file, as well as any necessary service imports from the server.  Sorry if this was unclear from the Requirements spec.

JonathanAldrich (changes requested):
A few comments to address--the major one is that we will need to return entire files to the client, not individual `service` declarations.  Sorry, this was probably not as clear as it should have been, although the example does show the client importing a service from the server which is the reason for this.

meerkat-lib/src/net/types.rs
```diff
@@ -166,6 +166,29 @@ pub enum MeerkatMessage {
...
+    /// #39: request the source code of a service class by name. A (browser)
+    /// client sends this to fetch a service it imports so it can parse and
+    /// instantiate the service locally. The service name is carried as a
+    /// string and resolved through the receiver's own interner, like the
+    /// other cross-node messages.
+    ServiceCodeRequest {
+        request_id: u64,
+        service: String,
+        reply_to: String,
+    },
```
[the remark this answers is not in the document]
ShreejithBabuG: Real point, but reply_to in the payload is the same pattern every existing message uses (LookupRequest, ActionRequest, RequestUpdates), tracked protocol-wide in #118. This PR length-limits it per review; binding replies to authenticated peer identity belongs in that protocol-wide change rather than one message.

meerkat-lib/src/net/types.rs
ShreejithBabuG: Done. Added codec::validate_service_code_request, which length-limits path and reply_to (MAX_NET_REQUEST_STRING_LENGTH). The server validates before responding.

meerkat-lib/src/net/types.rs
ShreejithBabuG: Right, the source is left unbounded; only path and reply_to are length-limited.

meerkat-lib/src/runtime/manager/mod.rs
ShreejithBabuG: Removed the duplicate service-creation loop. fetch_service_source now just returns the whole file; processing it (services and imports) through the normal program-loading path is done by the caller, which is wired in the browser client PR. Happy to extract that loop into a shared loader now if you would prefer.

meerkat-lib/src/runtime/parser/mod.rs
ShreejithBabuG: Done. Removed the per-service extraction entirely; the server now returns the whole .mkt file so imports and any other services come with it.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -1114,6 +1123,71 @@ impl Manager {
...
+            if let crate::ast::Stmt::Service { name, decls } = stmt {
+                self.create_service(*name, decls.clone())
+                    .await
+                    .map_err(|e| EvalError::RemoteDispatchFailed(e.to_string()))?;
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
ShreejithBabuG: Superseded by the redesign — the instantiate loop was removed in favor of reusing the normal program-loading path, so this no longer applies.

meerkat-lib/src/runtime/parser/mod.rs
```diff
@@ -201,6 +201,51 @@ fn parse_expr_fragment(src: &str, interner: &mut Interner) -> Result<crate::ast:
...
+        let boundary = source[after..]
+            .chars()
+            .next()
+            .map(|c| !c.is_alphanumeric() && c != '_')
+            .unwrap_or(true);
+        if !boundary {
+            search_from = after;
+            continue;
+        }
+        let brace_rel = source[start..].find('{')?;
```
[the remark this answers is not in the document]
ShreejithBabuG: Superseded — the per-service slicing was removed in favor of returning the whole file, so the boundary check no longer applies.

meerkat/src/main.rs
```diff
@@ -254,11 +254,15 @@ fn listen_success_addr(reply: NetworkReply) -> Result<Address, Box<dyn Error>> {
...
     prog: Vec<Stmt>,
+    input_file: &str,
     remote_url_map: std::collections::HashMap<String, String>,
     port: u16,
     local: bool,
     interner: Interner,
 ) -> Result<(), Box<dyn Error>> {
+    // #39: keep the raw source so we can answer ServiceCodeRequest by slicing
+    // out the requested service's `service <name> { ... }` block on demand.
+    let server_source = std::fs::read_to_string(input_file).unwrap_or_default();
```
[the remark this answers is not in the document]
ShreejithBabuG: Done. run_server now propagates the file-read error via ? instead of unwrap_or_default.

meerkat-lib/src/runtime/manager/mod.rs
JonathanAldrich: move this comment to be next to the code that goes with it

meerkat/src/main.rs
```diff
@@ -582,25 +585,24 @@ async fn run_server(
...
                 }
-                // #39: a client is asking for a service's source code by name.
-                // Slice the requested service's block out of the server source
-                // and reply with it, or reply with an error if not found.
+                // #39: a client is requesting a .mkt file by path. Return the
+                // whole file source (via the shared codec helper, which also
+                // validates the length-bounded fields) so the client can
+                // process it (services and any imports) through the normal
+                // program-loading path. The server currently hosts one program;
+                // mapping multiple paths to different files is a future
```
JonathanAldrich: PR 3 of 3 won't actually work unless you actually load the file in the path!  Because the client will need to run different code than the server in any real application, and it looks like this code ignores the path and returns the server code.  Anyway it is OK to postpone this to the next PR, or do it now, but it does need to be in this PR sequence, otherwise there is little point (hard even to test).

JonathanAldrich (changes requested):
A couple small points, plus what CoPilot identified.  Mostly about documentation and what needs to happen in the next PR.

meerkat/src/main.rs
```diff
@@ -254,11 +254,18 @@ fn listen_success_addr(reply: NetworkReply) -> Result<Address, Box<dyn Error>> {
...
+    input_file: &str,
     remote_url_map: std::collections::HashMap<String, String>,
     port: u16,
     local: bool,
     interner: Interner,
 ) -> Result<(), Box<dyn Error>> {
+    // #39: keep the raw source so we can answer ServiceCodeRequest by slicing
+    // out the requested service's `service <name> { ... }` block on demand.
+    // Propagate a read error rather than silently serving empty source, which
+    // would make every code request report "not found".
```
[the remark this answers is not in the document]
ShreejithBabuG: Fixed — updated to describe returning the whole .mkt program text, matching the implementation.

meerkat/src/main.rs
```diff
@@ -578,6 +585,32 @@ async fn run_server(
...
+                        server_source.clone(),
+                    );
+                    if let Some(net) = manager.network.as_mut() {
+                        net.handle_command(NetworkCommand::SendMessage {
+                            addr: Address::new(&reply_to),
+                            msg: response,
+                        })
+                        .await;
+                    }
+                }
```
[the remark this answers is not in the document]
ShreejithBabuG: The builder now takes &str and allocates the source only after validation passes, so a rejected request does no extra work.

meerkat/src/main.rs
```diff
@@ -578,6 +585,32 @@ async fn run_server(
...
+                MeerkatMessage::ServiceCodeRequest {
+                    request_id,
+                    path,
+                    reply_to,
+                } => {
+                    let response = codec::build_service_code_response(
+                        request_id,
+                        path,
+                        &reply_to,
+                        server_source.clone(),
```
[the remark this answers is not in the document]
ShreejithBabuG: The whole-file return is intentional per the review redesign, and extract_service_source was removed. Updated the PR description to match. Note: the server will additionally resolve the path to the requested file, landing in this PR sequence.

meerkat-lib/src/net/types.rs
```diff
@@ -166,6 +166,29 @@ pub enum MeerkatMessage {
...
+    /// path. The source is unbounded in length; only the request path and
+    /// reply address are length-limited.
+    ServiceCodeResponse {
+        request_id: u64,
+        path: String,
+        source: String,
+    },
+
+    /// #39: the requested service class was not found on this server.
+    ServiceCodeError { request_id: u64, error: String },
```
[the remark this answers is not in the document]
ShreejithBabuG: Broadened the doc to cover invalid-request (oversized path/reply_to) as well as missing-resource errors.

meerkat-lib/src/runtime/manager/mod.rs
```diff
@@ -1114,6 +1123,52 @@ impl Manager {
     ///
     /// Raises:
     ///     EvalError::LocalDispatchFailed: If a timeout or dispatch error occurs
+    /// #39: Fetch the source code of a service from a remote server by name.
+    ///
+    /// Sends a `ServiceCodeRequest` and awaits the reply, reusing the same
+    /// request/reply machinery as remote lookups. Returns the service's source
+    /// text of the requested `.mkt` file. The caller processes it through the
```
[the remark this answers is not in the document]
ShreejithBabuG: Fixed — the doc now says the file is fetched by path, matching the path-based API.

meerkat/src/main.rs
ShreejithBabuG: Implemented in this PR rather than deferring it. The server now resolves the requested path against its configured base directory and returns the contents of that file, allowing clients to request code independently of the server's own source. resolve_served_file serves as a zero-trust boundary: it rejects absolute paths and canonicalizes requests to prevent **..** and symlink traversal. Unit tests cover valid requests, absolute-path rejection, traversal attempts, and missing files. The round-trip test writes a real file to disk and verifies that it is served correctly.

meerkat-lib/src/runtime/manager/mod.rs
ShreejithBabuG: Done. I moved the remote_lookup documentation comment back above its corresponding function after it had drifted when fetch_service_source was inserted.

ShreejithBabuG:
This is the same systemic reply_to return-route issue tracked in #118. The
file-serving path makes the impact more significant since it can reflect an
entire file rather than a single value, but the correct fix is still to
bind replies to the authenticated transport identity as part of the
protocol-wide change. I've updated the comment to reference #118 instead of
#80/#111, since those track the interner work.

> Message ID: ***@***.***>
>

-- 
::DISCLAIMER::

---------------------------------------------------------------------
The 
contents of this e-mail and any attachment(s) are confidential and
intended 
for the named recipient(s) only. Views or opinions, if any,
presented in 
this email are solely those of the author and may not
necessarily reflect 
the views or opinions of SSN Institutions (SSN) or its
affiliates. Any form 
of reproduction, dissemination, copying, disclosure,
modification, 
distribution and / or publication of this message without the
prior written 
consent of authorized representative of SSN is strictly
prohibited. If you 
have received this email in error please delete it and
notify the sender 
immediately.
---------------------------------------------------------------------
Header of this mail should have a valid DKIM signature for the domain 
ssn.edu.in <http://www.ssn.edu.in/>

JonathanAldrich (approved):

Issue #95, opened 2026-06-24
REPL: allow import to accept a URL or file path


JonathanAldrich:
In the REPL, `import <service-name>` will load and run the service with the name given that is defined in the file `service-name.mkt` from the current directory.  This is nice but limited; what if the service we want to access is somewhere on the network, or in a file not in the current directory?  We want to generalize what you can put in an import statement to include file pathnames and URLs.

`import <service-name> from <URL>` should connect to a remote service, and work just like the `-i` command line flag.  It might be easiest to pub the <URL> part in a string quote; I am open to doing that or not.

`import <service-name> from <path>` should do the same thing as the current REPL import statement, but use the relative or absolute path `<path>` to find the file that defines the service. It is an error if the file is missing, cannot be read, is invalid Meerkat code, or does not contain a `service` definition with the name `<service-name>`.

PR #165, opened 2026-07-19
feat: allow import to accept a URL or file path

xcatita:
Added the possibility of importing a service with new grammar to be able to import a service with an absolute or relative path, or a URL to the service. Closes issue #95.
AST was changed to allow easy differentiation between an import statement referencing a file in the current directory (base directory in the case of the main's import statement handling) and one that uses a file path. This necessitated some changes to `AstPrinter` and other relevant function signatures. 
Lalrpop file was changed to accept the new type of import statement described in the issue, `import <service-name> from "<path/URL>"`.
Import statement execution was altered in the REPL to support both the old import statements and the new, which behave as described in issue #95. The REPL's network initializer was made to always run so URL imports are supported even if `remote_url_map` is empty. 
This was mostly copied over to the main's `run_client` function for consistency, except for URL imports.

## Summary by CodeRabbit

* **New Features**
  * Added support for explicit-path imports (e.g., `import Service from "./path.mkt"`).
  * Explicit-path imports now load only the named service and show clear single-service status.
  * Remote service imports can use direct network addresses and improve import status reporting.
* **Bug Fixes**
  * REPL network initialization now proceeds consistently and continues without network support if it fails.
  * Explicit-path imports are resolved from files without unintentionally switching to remote resolution.
* **Tests**
  * Added coverage for explicit-path import behavior and member access.

### DIFF
--- a/meerkat/src/repl.rs
+++ b/meerkat/src/repl.rs
@@ -249,17 +249,13 @@ async fn exec_stmt(
                     svc_name_str, address
                 )));
             }
-            if let Some(output) = crate::import_from_file(
+            crate::import_from_file(
                 manager,
                 explicit_path,
                 std::path::PathBuf::from(&path),
                 service_name,
             )
-            .await?
-            {
-                return Ok(Some(output));
-            }
-            Ok(None)
+            .await
         }
         Stmt::ActionStmt(action_stmt) => {
             let effect = execute(&action_stmt, repl_env, manager, Symbol::empty(), None)
--- a/meerkat/src/repl.rs
+++ b/meerkat/src/repl.rs
@@ -1,8 +1,8 @@
 use meerkat_lib::runtime::ast::{Expr, Stmt, Value};
 use meerkat_lib::runtime::interner::Symbol;
 use meerkat_lib::runtime::interpreter::{eval, execute, EvalContext, ExecuteEffect};
+use meerkat_lib::runtime::parser::parse_repl;
 use meerkat_lib::runtime::parser::ReplParseResult;
-use meerkat_lib::runtime::parser::{parse_file, parse_repl};
 use meerkat_lib::runtime::Manager;
 
 use directories::ProjectDirs;
@@ -249,38 +249,17 @@ async fn exec_stmt(
                     svc_name_str, address
                 )));
             }
-            let import_stmts = parse_file(&path, &mut manager.interner)
-                .map_err(|e| format!("Import '{}': {}", path, e))?;
-            let mut services = import_stmts.into_iter().filter_map(|stmt| match stmt {
-                Stmt::Service { name, decls } => Some((name, decls)),
-                _ => None,
-            });
-            if explicit_path {
-                if let Some((name, decls)) = services.find(|(name, _)| *name == service_name) {
-                    manager.create_service(name, decls).await.map_err(|e| {
-                        format!("Imported service '{}': {}", manager.interner.get(name), e)
-                    })?;
-                    return Ok(Some(format!(
-                        "Imported service: {}.",
-                        manager.interner.get(service_name)
-                    )));
-                } else {
-                    return Err(format!(
-                        "Service '{}' not found in '{}'",
-                        manager.interner.get(service_name),
-                        path
-                    )
-                    .into());
-                }
+            if let Some(output) = crate::import_from_file(
+                manager,
+                explicit_path,
+                std::path::PathBuf::from(&path),
+                service_name,
+            )
+            .await?
+            {
+                return Ok(Some(output));
             }
-            let mut loaded = Vec::new();
-            for (name, decls) in services {
-                manager.create_service(name, decls).await.map_err(|e| {
-                    format!("Imported service '{}': {}", manager.interner.get(name), e)
-                })?;
-                loaded.push(manager.interner.get(name).to_string());
-            }
-            Ok(Some(format!("Imported service(s): {}.", loaded.join(", "))))
+            Ok(None)
         }
         Stmt::ActionStmt(action_stmt) => {
             let effect = execute(&action_stmt, repl_env, manager, Symbol::empty(), None)
--- a/meerkat/src/repl.rs
+++ b/meerkat/src/repl.rs
@@ -79,7 +79,8 @@ pub async fn run_repl(
         println!();
     }
 
-    if !remote_url_map.is_empty() {
+    {
+        //Network initializer always runs
         let mut n = meerkat_lib::net::NetworkActor::new(meerkat_lib::net::types::NodeType::Server)
             .await
             .map_err(|e| format!("Network error: {}", e))?;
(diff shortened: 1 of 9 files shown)


meerkat/src/repl.rs
```diff
@@ -217,28 +227,59 @@ async fn exec_stmt(
...
-                    svc_name_str, url
+                    svc_name_str, address
                 )));
             }
             let import_stmts = parse_file(&path, &mut manager.interner)
                 .map_err(|e| format!("Import '{}': {}", path, e))?;
-            let mut loaded = Vec::new();
-            for s in import_stmts {
-                if let Stmt::Service { name, decls } = s {
+            let mut services = import_stmts.into_iter().filter_map(|stmt| match stmt {
```
JonathanAldrich: Looks like lines 254-282 are mostly the same as similar code in `main.rs`. Is it possible to eliminate this duplication?

(I recognize that some of the duplication might be pre-existing, but it would be better to fix it, if possible, than to make it worse)

JonathanAldrich (changes requested):
Overall this works well for me and will be a nice enhancement!  I had one request regarding duplicate code, can you have a look?

meerkat/src/repl.rs
xcatita: You're right; will do.

xcatita:
@coderabbitai review

JonathanAldrich:
@xcatita can you have a look at pull request 170?

https://github.com/meerkat-lang/meerkat/pull/170

As part of finishing the implementation of static checking, Dustin has been working on changing the architecture of imports a bit.  It would be great to fit in your extension with that work, so any feedback you have on it would be valuable.

We will also have to resolve any conflicts between the two PRs; with my apologies I am leaning towards landing 170 first just because it makes architectural improvements on which your extensions could build.  I hope that's OK and won't result in a lot of extra work for you.

xcatita:
Of course, no apologies necessary; thanks for the heads-up. I will likely not be available during the next week to address these changes, but afterwards I will look into adapting my work to the new architecture.

JonathanAldrich:
Ok!  I merged PR 170 so whenever you have time you can look at updating this.  I appreciate your patience, version control resolves many things but not all.

xcatita:
@coderabbitai review

Cargo.lock
dustinjuliano, line 1: Recommend not including changes to `Cargo.lock` in this PR and keeping module or package updates as separate, atomic commits. It appears as though `cargo update` was ran and this file was included in a git all file commit. I would propose a fixup commit that restores it to match `main` branch.

dustinjuliano:
A request was made today (2026-08-24) in the research group Slack by @xcatita for RFC and/or assistance/review on this PR. Since I designed and implemented the import system I am volunteering to give a hand. Please note that I may be delayed in some responses as I have returned to my home institution and have a full load of courses plus new external research projects.

My first task was to clone this PR and do an analysis. Here is a very brief overview of what I found:

The grammar addition in `meerkat.lalrpop` and AST definition in `ast/mod.rs` are straightforward. However:

1. Remote service registration may have been severed in `run_client` L1044-L1120 by eliminating `manager.remote_services` mapping.
2. There is an uncoordinated, redundant second import pass inside `run_client` that executes synchronously and drops network commands.
3. There is double-parsing and duplicate validation inside `imports.rs` L190-L425.
4. The import state machine appears bypassed in `repl.rs` L80-L130 with a custom loader that ignores transitive dependencies.
5. There may be misinterpretation of `Manager::create_service` and `unified_ast` ownership semantics, which could affect AST cloning and mutation.

Those issues are related to the next finding, which is that it looks like this PR was not rebased against `main` from PR 170. It appears that attempts were made to manually merge changes in one or more merge commits. This is shown from git diagnostics showing the parent branches. This may be the root cause of the difficulties.

PR 170 was very large. This PR (165) began shortly after or before some of the major changes and design were happening on PR 170. PR 165 was given notice of this and went dormant for a time after some work had been done. PR 170 went forward in full force and other PRs were deferred to rebase against it or the branch I created for it.

Merge commits were performed. However, if a rebase would have been done against PR 170 the merge conflicts would have shown that significant parts of PR 165 would be incompatible with the existing architecture. There is a high probability of having to delete or rework some things. The good news, however, is that the actual changes from PR 165 to realize local path resolution will be greatly simplified because a lot of the needed infrastructure was added in PR 170 due to the overloaded semantics of service names mapping to the file system.

The import system is highly sophisticated, and I also would have liked to have more time to improve it. We made intentional tradeoffs for simplicity over efficiency. That means that some copying of the AST is acceptable; however, this should not be required to add paths to the import semantics, as it is just telling the existing file import resolution code paths to use a specific path name instead of the implicit one given by the service name itself. It should not be required to manipulate the ASTs, as that is already implemented.

There are also significant downstream effects from the import system and how it had to be fit into a larger puzzle piece of the live code updates and the static semantics, and none of these, unfortunately, can be altered without affecting the entire pipeline. This is why I would encourage trying to keep the alterations as localized as possible.

Here is an outline of what I would recommend. You, of course, do not have to do this. It is just me trying to be helpful:

- Copy the current branch to a backup branch (`git branch backup-pr-165`). Do not skip this step because the following steps are destructive.
- Perform `git reset --hard origin/main` on your PR branch.
- Copy over the grammar and AST changes from commit `ad4fd33` (in `meerkat.lalrpop`, `ast/mod.rs`, `printer.rs`, and pattern match arms).
- Test and make an atomic commit for just those grammar/AST changes. This establishes a clean baseline on top of `main`.
- Do surgical updates to `imports.rs` for explicit path resolution:
  - In `resolve_import`: Use `base_dir.join(target_path)` when `explicit_path == true` instead of `base_dir.join(format!("{}.mkt", service_name))`.
  - In `on_recv_source`: Parse once (`parser::parse_string`) and verify the specified service declaration exists in the AST. Avoid reading or parsing the file multiple times.
  - Leave `main.rs` and `repl.rs` untouched. `Node::resolve_imports_with_net` and `Imports` already handle file loading and static checks before runtime execution.
- Carry over the integration tests (`test_imports_with_explicit_path` from `b1ec6fd` and `test_integration_import_path_member_access` from `124eab7`).
- Keep `Cargo.lock` untouched and identical to `main`.

This should help get PR 165 on track with minimal damage or time lost, as you have example code and an outline to help guide. If you run into any trouble or have any further questions just let me know. I do apologize for the overlap in work; however, the architectural vision for PR 170 was set out in advance and was part of a larger plan to enable type-safe live code updates and had to displace several other tracks at the time.

dustinjuliano:
@xcatita Reached out again and had further questions. I am sharing them here so they can be used for future reference. I had to interpret their question multiple ways, so I expanded them into four possible meanings of the original question from Slack. Let me know if this does not cover the intended meaning.

Question 1: Why `run_client` doesn't just use `full_ast` to create services?

The variable `full_ast` contains remote services. During the static phase, `Node::resolve_imports_with_net` fetches the ASTs of remote services so it can typecheck calls into them.

If `run_client` just looped through `full_ast` and called `create_service`, the client would locally instantiate remote services.

The `unified_ast` variable is a reference AST that is stored on `manager` so the runtime knows the type signatures for live code updates (`Stmt::Update` / `Stmt::Atomic`), not as a list of local processes to spin up.

Question 2: Why does `run_client` re-reads and re-parses the local file in `Stmt::Import`?

This was an intentional design tradeoff to maximize simplicity over caching. The static phase (`Node`) resolves the dependency graph and builds `unified_ast` for type-checking. The runtime runner (`run_client`) walks `prog` (the entry file). When it sees `Stmt::Import`, it checks if is in `remote_url_map`? If yes, it registers the remote address in `manager.remote_services`. Otherwise it checks if it is a local file. If yes, it loads the service into `manager`.

The simplicity of re-parsing the local file at startup keeps `run_client` completely decoupled from having to reconstruct runtime bindings out of `unified_ast`. This was a directive given to me by Professor Aldrich to keep things as simple as possible so we could maintain forward momentum.

Question 3: Does `run_client` need custom filtering for `import s1 from "file.mkt"`?

I believe the answer is that it does not. We should not assume that if `file.mkt` has both `s1` and `s2`, `run_client` must filter and only instantiate `s1`. In Meerkat, files are kinda like compilation units; when a file is loaded, its services are created so dependencies can be satisfied. The validation that `s1` actually exists in that file belongs in **`imports.rs`** during the static phase (`on_recv_source`), not as filtering in `run_client`.


Question 4: Does `exec_stmt` in `repl.rs` need custom logic to execute explicit path imports, or does existing code on `main` already handle them?

I believe in `main`, `exec_stmt` already does this:

  ```rust
  Stmt::Import { path, service_name, .. } => {
      let import_stmts = parse_file(&path, ...);
      ...
  }
  ```

Because the grammar sets `path` to the explicit string literal, `parse_file(&path)` should already work for explicit paths. So it seems that what `repl.rs` needs is updating the pattern match to ignore `explicit_path: _`. The custom loader was not needed because of additions made in PR 170 and was dropping transitive dependencies.

xcatita:
@dustinjuliano Thank you for the clarification, I do believe it's enough for me to start readying the PR for review again.

xcatita:
@coderabbitai review

PR #168, opened 2026-07-20
feat: vector clock for glitch-free updates

andrewxzhou:
### Description
Implements the vector clock mechanism for glitch-free updates as described in the Meerkat semantics document. Essentially, each variable and definition holds a `vector_clock : VClock` which tracks how many updates that variable has seen. Thus, a `VClock` is a type alias for a `HashMap<(Symbol, Symbol), u64>`, where keys are `(svc, var)` tuples and values are a counter for how many times that variable has been updated. By convention a missing entry is equivalent to a clock count of 0. 

### Checklist
[x] Make vector clock type
[x] Update vector clock in the places where it should be updated for local updates
[x] Update vector clock in the places where it should be updated for remote updates
[x] write tests for local updates
[x] write tests for remote updates
[ ] maybe write some more tests

### Additional Info/Notes for review
I did not write any test cases by hand 👍 but I will review them to make sure they're not doing anything stupid. They call me a poly-time decider......

## Summary by CodeRabbit

- **New Features**
  - Added vector-clock tracking to updates, transactions, and dependency state.
  - Network updates now carry synchronization metadata for consistent distributed reactivity.
  - Added encoding and decoding support for vector clocks.

- **Bug Fixes**
  - Improved handling of concurrent and dependent updates, including diamond-shaped workflows.
  - Added validation to reject malformed synchronization data.
  - Improved type checking for tuple parameters and function calls.


andrewxzhou:
AI review comments have been addressed. Should be ready for human review?

andrewxzhou:
okay it seems like there's something weird happening with the vector clock gate. From what I can tell the copilot suggestion on R711 doesn't work, but I'm not entirely sure why. currently investigating, will update here with more findings as they come up

